PR #6, opened 2019-04-05
Update IAM policy member domain constraint

t12g:
To optionally reject "project reference" members (example: "projectEditor:my-project"). They are allowed by default.


validator/test/fixtures/constraints/iam_allowed_policy_member_reject_project_reference/data.yaml
```diff
@@ -0,0 +1,15 @@
...
+  severity: high
+  match:
+    target: ["organization/*"]
+    exclude: [] # optional, default is no exclusions
+  parameters:
+    allow_project_references: false
+    domains:
+      - gserviceaccount.com
+      - google.com
+      - notgoogle.com
```
alexsgoogle: Need to include projectViewer or something here?

validator/test/fixtures/constraints/iam_allowed_policy_member_reject_project_reference/data.yaml
```diff
@@ -0,0 +1,15 @@
...
+kind: GCPIAMAllowedPolicyMemberDomainsConstraint
+metadata:
+  name: reject_project_reference
+spec:
+  severity: high
+  match:
+    target: ["organization/*"]
+    exclude: [] # optional, default is no exclusions
+  parameters:
+    allow_project_references: false
```
joecheuk: is project reference the right name for those iam bindings? doesn't sound intuitive from a user point of view.

t12g: good point. I looked at https://cloud.google.com/storage/docs/access-control/iam#acls, and it just refers to them as some specific member types that are different from the typical ones. 
But that got me thinking that instead of using boolean parameter, we could use a whitelist of IAM member types. That's more explicit and flexible. I've made this change; PTAL.

validator/test/fixtures/constraints/iam_allowed_policy_member_reject_project_reference/data.yaml
t12g: This is now outdated. Previously it was controlled by a boolean parameter. Now we're using a whitelist of member types to ignore.

morgante:
Should we add a test for the default behavior (whitelisted) on top of the existing one which has an empty whitelist?

Otherwise LGTM.

t12g:
> Should we add a test for the default behavior (whitelisted) on top of the existing one which has an empty whitelist?
> 
> Otherwise LGTM.

I think the default behavior w/o explicit whitelist should be covered by the two existing test constraints:
https://github.com/forseti-security/policy-library/tree/master/validator/test/fixtures/constraints/iam_allowed_policy_member_two_domains
https://github.com/forseti-security/policy-library/tree/master/validator/test/fixtures/constraints/iam_allowed_policy_member_all_domains

morgante:
Got it, yes LGTM.

PR #11, opened 2019-04-12
Add cloudbuild support

t12g:
It just runs "make test" for now.
See https://github.com/t12g/policy-library/pull/2/checks?check_run_id=99625649 for test run output.


cloudbuild/Dockerfile
```diff
@@ -0,0 +1,11 @@
+# This Dockerfile builds the image used in Cloud Build CI to run 'make test'.
+# To push a new image, run 'gcloud builds submit --project=config-validator --tag gcr.io/config-validator/make .'
```
morgante: Can we add a `make` target for building this?

t12g: Sure. I'll do that in a follow-up PR.

PR #19, opened 2019-04-20
Add version information as constraint template name suffix

t12g:
As discussed, we'll start with v1. If this looks good, I'll apply the change to other constraint templates in a follow-up PR.

Also remove the following fields:
* "listKind", since it defaults to <kind>List
* "constraintVersion", since version it's now part of constraint template name

### DIFF
--- a/policies/templates/gcp_always_violates.yaml
+++ b/policies/templates/gcp_always_violates.yaml
@@ -20,7 +20,7 @@ spec:
   crd:
     spec:
       names:
-        kind: GCPAlwaysViolatesConstraintv1
+        kind: GCPAlwaysViolatesConstraintV1
         plural: GCPAlwaysViolatesConstraints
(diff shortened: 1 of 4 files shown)


policies/templates/gcp_always_violates.yaml
```diff
@@ -20,8 +20,7 @@ spec:
   crd:
     spec:
       names:
-        kind: GCPAlwaysViolatesConstraint
-        listKind: GCPAlwaysViolatesConstraintList
+        kind: GCPAlwaysViolatesConstraintv1
```
morgante: I think capitalizing V will improve readability.

```suggestion
        kind: GCPAlwaysViolatesConstraintV1
```

t12g: Done.

PR #22, opened 2019-04-23
Append "V1" version suffix to constraint template name

t12g:
Follow up to https://github.com/forseti-security/policy-library/pull/19

Also remove redundant listKind field in constraint templates.

### DIFF
--- a/policies/templates/gcp_external_ip_access.yaml
+++ b/policies/templates/gcp_external_ip_access.yaml
@@ -23,8 +23,7 @@ spec:
   crd:
     spec:
       names:
-        kind: GCPExternalIpAccessConstraint
-        listKind: GCPExternalIpAccessConstraintList
+        kind: GCPExternalIpAccessConstraintV1
         plural: GCPExternalIpAccessConstraints
         singular: GCPExternalIpAccessConstraint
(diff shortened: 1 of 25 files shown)


policies/templates/gcp_external_ip_access.yaml
```diff
@@ -23,8 +23,7 @@ spec:
   crd:
     spec:
       names:
-        kind: GCPExternalIpAccessConstraint
-        listKind: GCPExternalIpAccessConstraintList
+        kind: GCPExternalIpAccessConstraintV1
         plural: GCPExternalIpAccessConstraints
```
morgante: Doesn't the plural also need to be versioned?

t12g: Done.

Also switched it to lowercase and removed singular field. 
https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.14/#customresourcedefinitionnames-v1beta1-apiextensions-k8s-io mentioned that plural must be all lowercase and that singular defaults to lowercased <kind>

t12g:
> Please also update the kind with version in the yaml files under samples/

Done.

PR #30, opened 2019-04-27
Add a section to go over authoring convention

t12g:
Also run md formatter.


docs/user_guide.md
```diff
@@ -528,14 +527,38 @@ This section is only applicable to advanced users who wish to create custom
...
 
-#### Template Naming Convention
+#### Template Authoring Convention
+
+##### Naming
+
+The template name appears in three places in a template YAML file:
+
+*   metadata name: All lower case with "-" as the separator. It has the format
+    of "gcp-{resource}-{feature}-{version}" (example: "gcp-storage-logging-v1").
```
t12g: Do we think the "gcp" prefix is needed here? It feels redundant to append a "gcp" prefix in front of names like "gce", "gke", or "cloudsql". 

We could adopt a convention where the "gcp" prefix is used if the constraint applies to multiple resources. Thoughts?

t12g: We don't have the version number as part of the metadata name. I think we need to make this change.

t12g: Is "resource" the right term to use? Any suggestions?

joecheuk: I do feel like gcp might be a bit redundant here, unless we plan to go with multiple cloud in the future then this would make sense, is that part of the plan?

joecheuk: I think resource is a good way of grouping them

docs/user_guide.md
```diff
@@ -528,14 +527,38 @@ This section is only applicable to advanced users who wish to create custom
...
+*   CRD kind: Camel case. It has the format of
```
joecheuk: should we be more specific when mentioning CRD kind and CRD plural name? kind and plural aren't located directly under crd -
crd:
  spec:
    names:
      kind:
      plural:

t12g: Done

docs/user_guide.md
t12g: That's a possibility. In that case, having a "gcp" prefix sounds preferable. 

But we still have this redundancy issue. One option is not use terms like "gce", "gcs", or "gke".
Instead, we'll use "gcp-compute", "gcp-storage", and "gcp-kubernetes" (or should it be "gcp-container-clusters" if we want to follow gcloud's convention?)

Thoughts?

morgante: i think having the `gcp` prefix across the board is helpful to differentiate from, at a minimum, Kubernetes constraints.

In terms of redundancy, I think `gcp-compute` and such (following gcloud conventions) make sense.

joecheuk: Do you guys think the file name might get too long if we put too much info there?

morgante: I don't think that's a major issue, but I also don't think this is actually an area where either decision has a big impact.

PR #38, opened 2019-05-03
add dnssec constraint template

tangyan:


validator/dnssec.rego
```diff
@@ -0,0 +1,33 @@
...
+
+deny[{
+	"msg": message,
+	"details": metadata,
+}] {
+	constraint := input.constraint
+	asset := input.asset
+	asset.asset_type == "dns.googleapis.com/ManagedZone"
+
+	asset.resource.data.dnssecConfig.state == "OFF"
```
t12g: can we compare it to not on instead? there's a 3rd "transfer" state

validator/test/fixtures/constraints/require_dnssec/data.yaml
```diff
@@ -0,0 +1,25 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+apiVersion: constraints.gatekeeper.sh/v1alpha1
+kind: GCPDNSSECEnabledConstraintV1
+metadata:
+  name: require_dnssec
+spec:
+  constraintVersion: 0.1.0
```
t12g: let's delete this line; it's no longer in spec

validator/dnssec_test.rego
```diff
@@ -0,0 +1,37 @@
...
+
+	violation := issues[_]
+}
+
+# Confirm total violations count
+test_dnssec_violations_count {
+	count(all_violations) == 1
+}
+
+test_dnssec_violations_basic {
```
t12g: can we add the sample data for the transfer case and check it?

tangyan: Done

validator/dnssec.rego
tangyan: Done

validator/test/fixtures/constraints/require_dnssec/data.yaml
tangyan: Done

PR #42, opened 2019-05-07
Update .rego file names

t12g:
To include "gcp" prefix and version suffix. 
Also:
* include "compute" in the external IP constraint name
* rename IAM domain constraint to just "GCPIAMAllowedDomainsConstraintV1"


morgante:
@t12g I'm not sure we need to include versioning in the actual Rego files. I don't expect us to have multiple versions simultaneously maintained in *this* repo. Likewise, I don't know if the GCP prefix is needed for the actual Rego files.

t12g:
@morgante could you clarify the not having "multiple versions simultaneously maintained in _this_ repo" part?

When we add Vn+1 template, we'll delete Vn as part of the commit?
Or, is that Vn will be "frozen" once Vn+1 is available. We won't change Vn anymore, and therefore the .rego file is always associated with the most recent version?

morgante:
@t12g I think we'd delete Vn as part of the commit adding Vn+1. Even though we allow customers to have both versions in their cluster simultaneously, I don't think we should continue to support and expose multiple versions.

t12g:
Clarified the versioning story in https://github.com/forseti-security/policy-library/pull/61.

PR #48, opened 2019-05-14
Updated the README to include a link to the constraint template authoring guide.

mmontan:


mmontan:
Fly-by correction to the documentation. I noticed that the link from the "create a new constraint" pointed to the user guide again. It was confusing when I was reading the documentation.

PR #53, opened 2019-05-15
Restrict CloudSQL instances with public IP

kevensen:
Initial commit of forbiding CloudSQL instances from having a public IP


joecheuk:
@morgante @t12g  based on our last discussion, it seems to be more natural to fit this to the gcp external ip access constraint template, is that still the case?

https://github.com/forseti-security/policy-library/blob/master/policies/templates/gcp_external_ip_access_v1.yaml

morgante:
@joecheuk I think that constraint should actually be specifically for VMs while we add this new CloudSQL one.

Eventually we could consider a super-constraint for all resource types.

validator/sql_public_ip.rego
```diff
@@ -0,0 +1,16 @@
+package templates.gcp.SQLPUBLICIPV1
```
morgante: Can you add the Google license header?

Also, package should match constraint name; ie. `GCPSQLPublicIpConstraintV1`.

validator/sql_public_ip.rego
```diff
@@ -0,0 +1,16 @@
...
+import data.validator.gcp.lib as lib
+
+deny[{
+	"msg": message,
+	"details": metadata,
+}] {
+	asset := input.asset
+	asset.asset_type == "sqladmin.googleapis.com/Instance"
+
+	asset.resource.data.settings.ipConfiguration.ipv4Enabled == true
```
morgante: Is the `ipConfiguration` object always present?

validator/test/fixtures/constraints/sql_public_ip/data.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: constraints.gatekeeper.sh/v1alpha1
```
morgante: Needs license.

validator/sql_public_ip.rego
kevensen: Yes

policies/templates/gcp_sql_public_ip_v1.yaml
```diff
@@ -0,0 +1,48 @@
...
+        kind: GCPSQLPublicIpConstraintV1
+        plural: gcpsqlpublicipconstraintsv1
+      validation:
+        openAPIV3Schema:
+          properties: {}
+  targets:
+   validation.gcp.forsetisecurity.org:
+      rego: |
+            #INLINE("validator/sql_public_ip.rego")
+           package templates.gcp.SQLPUBLICIPV1
```
morgante: Please run `make build` again.

PR #56, opened 2019-05-15
Add policy to enforce CMEK encryption on buckets

aolarte:


policies/templates/gcp_storage_cmek_encryption_v1.yaml
```diff
@@ -0,0 +1,78 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+apiVersion: templates.gatekeeper.sh/v1alpha1
+kind: ConstraintTemplate
+metadata:
+  name: gcp-storage-cmek_encryption-v1
```
morgante:
```suggestion
  name: gcp-storage-cmek-encryption-v1
```

validator/storage_cmek_encryption.rego
```diff
@@ -0,0 +1,46 @@
...
+		"default_kms_key_name": kms_key_name,
+		"resource": asset.name,
+	}
+}
+
+###########################
+# Rule Utilities
+###########################
+default_kms_key_name(bucket) = default_kms_key_name {
+	encryption := lib.get_default(bucket, "encryption", "default")
```
morgante: Since we are looking at encryption as an object on line 76, default value should be an object.

```suggestion
	encryption := lib.get_default(bucket, "encryption", {})
```

validator/storage_cmek_encryption_test.rego
```diff
@@ -0,0 +1,37 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+package templates.gcp.GCPStorageCMEKEncryptionConstraintV1
+
+all_violations[violation] {
+	resource := data.test.fixtures.assets.storage_buckets[_]
+	constraint := data.test.fixtures.constraints.require_storage_logging
```
morgante: We should add a new constraint fixture specific to this rule, since we're not looking for the storage logging constraint.

Issue #63, opened 2019-05-18
Modifying fixture library likely to affect multiple constraint templates


ddremund:
It appears that as the test fixtures are maintained today, the typical constraint template test often includes counting "denies" produced against the entirety of the asset fixtures for a particular resource type or set of resource types.

It seems likely that in many cases, producing new asset fixtures as required for testing newly-developed policies has the side effect of also modifying the fixture library used for testing existing policies, which may break their tests, requiring the developer of a new constraint to either spend time crafting fixtures that meet previous expectations of other constraint tests, or modify those tests (from other fixtures than they were working on in their branch) to account for the newly added fixtures.

Should we consider changing the paradigm such that there is a default set of fixtures, and then perhaps a subdirectory for constraint-specific fixtures? I considered implementing this myself for a policy PR I recently submitted, but thought maybe it should be raised as an issue first.

t12g:
Thanks for bringing this up. Sharing test data between different constraint templates doesn't seem sustainable to me either. I'd suggest taking it one step further - is there a strong reason to have a common set of test fixtures? Maybe we should make every constraint template test to have its own private set of test data?

There will be some duplication, but it's just data, not logic. The benefit is that we'll never have to worry about whether modifying a test fixture will break other tests. Thoughts? @morgante - I'd like to get your take on this as well.

ddremund:
@t12g You're right, there would be some duplication but it is just data. I think that would vastly simplify things.

morgante:
> Maybe we should make every constraint template test to have its own private set of test data?

I agree, I'd suggest we change the fixtures to be constraint-specific.

ddremund:
It looks like some authors are creating their own fixture sets, some are using existing ones, and some are mixing the two. A standardized directory structure would be desirable, e.g. fixtures/<constraint-name>/<asset-type> or fixtures/<asset-type>/<constraint-name>. I prefer the former, myself.

morgante:
I think `fixtures/<asset-type>/<constraint-name>` is preferable.

ddremund:
I'm cool with `fixtures/<asset-type>/<constraint-name>`. I guess the good thing about that is it makes it easy for us to pre-create a directory structure for folks to drop their stuff in.

Once my two outstanding PRs are in I'll see what I can do with regard to creating the structure and patching some of the existing constraint tests.

morgante:
Excellent, thanks.

abombss:
[...]

PR #76, opened 2019-05-31
Adding GCPResourceValuePatternConstraintV1

abombss:
Adding a fairly generic constraint template that checks any field on an asset against a regex pattern for validation.  It started simply as a way to check labels, then evolved as we started writing more policies and needed to check different values and attributes.


validator/resource_value_pattern.rego
```diff
@@ -0,0 +1,151 @@
...
+	constraint := input.constraint
+	params := lib.get_constraint_params(constraint)
+	asset := input.asset
+
+	# asset type can be a scalar string or an array, only continue if the
+	# resource's asset type is one that should be looked at
+	# If asset types don't match exactly they will just be skipped
+	# potentially creating false positives.
+	# Ensure your assets types are 100% correct with negative testing.
+scalar_in(	asset.asset_type, params.asset_type)
```
t12g: since params.asset_type is an array, do we really need the `scalar_in` function? would just `asset.asset_type == params.asset_type[_]` be sufficient?

validator/resource_value_pattern.rego
```diff
@@ -0,0 +1,151 @@
...
+
+	rule_data := {
+		"is_required": lib.get_default(params, "required", true),
+		"has_field": has_field_by_path(asset.resource.data, field_name),
+		"has_pattern": lib.has_field(params, "pattern"),
+		"pattern": lib.get_default(params, "pattern", ""),
+		"value": get_default_by_path(asset.resource.data, field_name, ""),
+	}
+
+	# The rule is truthy so we ensure the asset is valid
```
t12g: nit: can we rework this comment a bit? I don't quite follow the "only violate if it is not as part of the deny rule" part.

validator/resource_value_pattern.rego
```diff
@@ -0,0 +1,151 @@
...
+is_required_field_valid(is_required, has_field) {
+	has_field == true
+}
+
+is_required_field_valid(is_required, has_field) {
+	is_required == false
+	has_field == false
+}
+
+is_pattern_valid(is_required, has_field, has_pattern, pattern, value) {
```
t12g: the `is_required` parameter doesn't seem to be used in `is_pattern_valid`?

validator/resource_value_pattern.rego
```diff
@@ -0,0 +1,151 @@
...
+get_field_by_path(obj, path) = output {
+	split(path, ".", path_parts)
+	walk(obj, [path_parts, output])
+}
+
+# internal helper because walk has strange behavior
+# such that we need to wrap it in order to get
+# explicit true / false
+_has_field_by_path(obj, path) {
+	_ := get_field_by_path(obj, path)
```
t12g: do we need the assignment here? would just `get_field_by_path(obj, path)` be sufficient?

validator/resource_value_pattern.rego
```diff
@@ -0,0 +1,151 @@
...
+
+has_field_by_path(obj, path) {
+	_has_field_by_path(obj, path)
+}
+
+has_field_by_path(obj, path) = false {
+	not _has_field_by_path(obj, path)
+}
+
+get_default_by_path(obj, path, _default) = output {
```
t12g: what's the significance of the underscore in front of `default`? that seems inconsistent from other parameters?

validator/resource_value_pattern.rego
```diff
@@ -0,0 +1,151 @@
...
+	true == has_field_by_path(obj, path)
```
t12g: is `true ==` necessary?

validator/test/fixtures/constraints/resource_value_pattern_required/data.yaml
```diff
@@ -0,0 +1,30 @@
...
+spec:
+  severity: high
+  match:
+    gcp:
+      target: ["organization/*"]
+      exclude: [] # optional, default is no exclusions
+  parameters:
+    asset_type: "cloudresourcemanager.googleapis.com/Project"
+    field_name: "labels.billing_id"
+    required: true
```
t12g: can we add 2 more test constraint fixtures: one with `required: false`, and the other with `required: true` and no pattern?

validator/resource_value_pattern.rego
abombss: @t12g Its modeled closely after the the existing lib rules of has_field and get_default just accepting a document and dynamic nested path using walk.  Walk has some strange behavior that requires we capture a variable in order to return a specific true / false.  There is a discussion on slack about it [here](https://openpolicyagent.slack.com/messages/C1H19LW4F/convo/C1H19LW4F-1552948594.244300/).

validator/resource_value_pattern.rego
abombss: Its modeled identical to the existing rules in lib.  Should I change or keep them?

https://github.com/forseti-security/policy-library/blob/0dfa9432ea261e7605146d03a64a490e860e0327/lib/util.rego#L38

policies/templates/gcp_resource_value_pattern_v1.yaml
```diff
@@ -0,0 +1,358 @@
...
+spec:
+  crd:
+    spec:
+      names:
+        kind: GCPResourceValuePatternConstraintV1
+        plural: gcpresourcevaluepatternconstraintsv1
+      validation:
+        openAPIV3Schema:
+          properties:
+            asset_type:
```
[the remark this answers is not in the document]
abombss: Addressed with latest commits.

validator/resource_value_pattern.rego
abombss: Addressed with latest commits.

validator/resource_value_pattern.rego
abombss: Addressed with latest commits.

validator/resource_value_pattern.rego
abombss: Addressed with latest commits.

validator/test/fixtures/constraints/resource_value_pattern_required/data.yaml
abombss: Addressed with latest commits.

validator/resource_value_pattern.rego
abombss: Addressed with latest commits.

validator/resource_value_pattern.rego
t12g: thanks. can we mention this in the comment?

validator/resource_value_pattern.rego
t12g: ah, never mind, `default` is a keyword, and that's the reason for the underscore prefix.

validator/test/fixtures/assets/projects/data.json
```diff
@@ -0,0 +1,74 @@
+[{
```
t12g: Can we rename the directory (`project`) after the constraint template name? The expectation is that we don't plan to share test fixtures across templates to minimize any cross-test dependency on fixture data.

validator/resource_value_pattern_test.rego
```diff
@@ -0,0 +1,142 @@
...
+
+	issues := deny with input.asset as resource
+		 with input.constraint as constraint
+
+	violation := issues[_]
+}
+
+# Confirm total violations count
+test_resource_value_pattern_required_violations_count {
+	count(all_required_violations) == 2
```
t12g: instead of relying on count, can we verify the resource field in violation details to make sure that we're getting the expected violations?

abombss: I updated the tests and added a couple more as well as asserting the resource names.

validator/test/fixtures/assets/projects/data.json
abombss: Reorganized the tests according to the discussion in #63

t12g:
Hi @abombss - is this PR ready for another go?

abombss:
Yup, all items from the review should be resolved.

validator/resource_value_pattern_test.rego
```diff
@@ -0,0 +1,187 @@
...
+	# casted as an array all evals of X[_] would fail.  Tested iterating sets in
+	# playground and they work fine, so I am not sure the problem here.
+	a := cast_array(data.violations)
+	i := a[_]
+	trace(sprintf("Violations %v", [i]))
+	name := i.details.resource
+}
+
+test_resource_value_pattern_optional_on_multiple_resources {
+	# Arrange
```
t12g: nit: are the "Arrange", "Act" and "Assert" comments useful? They don't seem to convey any additional information?

validator/resource_value_pattern_test.rego
```diff
@@ -0,0 +1,187 @@
...
+	}
+
+	# Act
+	found_violations := find_violations with data.assets as fixture_assets
+		 with data.test_constraints as lookup_constraint.optional_billing_id_on_multiple_assets
+
+	found_resources := resource_names with data.violations as found_violations
+
+	# Assert
+	# opa eval -d=lib/ -d=validator/ --format=pretty -explain full "data.templates.gcp.GCPResourceValuePatternConstraintV1.test_resource_value_pattern_optional_on_multiple_resources"
```
t12g: nit: missing an extra dash before the `explain` flag

validator/resource_value_pattern_test.rego
```diff
@@ -0,0 +1,187 @@
...
+trace(	sprintf("Found violations count should be %v and got %v", [count(expected_resources), count(found_violations)]))
```
t12g: instead of count, should we trace `found_resources` and `expected_resources` instead? that'll tell us exactly what the difference is, right?

validator/resource_value_pattern_test.rego
```diff
@@ -0,0 +1,187 @@
...
+	asset := data.assets[_]
+	constraint := data.test_constraints[_]
+	issues := deny with input.asset as asset with input.constraint as constraint
+	total_issues := count(issues)
+	violation := issues[_]
+}
+
+# Helper to create a set of resource names from violations
+resource_names[name] {
+	# Not sure why I need this, data.violations was a array_set but unless
```
t12g: I'm not able to repro this issue. Maybe it's an OPA versioning thing? I'm using OPA 0.11.

alternatively, we can use a function for this:
```
found_resources(violations) = output {
	output := {x | x := violations[_].details.resource}
}
```

that seems more direct?

validator/resource_value_pattern_test.rego
```diff
@@ -0,0 +1,187 @@
...
+			"tf_foo": "foo/project/google/v123",
+		},
+		"name": "myname",
+	}
+
+	get_default_by_path(doc, "name", "foo") == "myname"
+	get_default_by_path(doc, "labels.id", "foo") == "1234"
+	get_default_by_path(doc, "labels.t", false) == true
+	get_default_by_path(doc, "labels.f", true) == false
+
```
t12g: add a comment here to denote that this block is verifying the default behavior? (or consider splitting this up to 2 tests)

AdrienWalkowiak:
@abombss are you still working on this PR?

ocervell:
@AdrienWalkowiak @abombss this is high value for a customer. Any chance to see this finished soon ?

AdrienWalkowiak:
@abombss I can take this one over if your plate is full

AdrienWalkowiak:
Closing - will reopen from my fork.

PR #77, opened 2019-05-31
Constraint template to verify audit logging enablement

t12g:


policies/templates/gcp_iam_audit_log.yaml
```diff
@@ -0,0 +1,92 @@
...
+      validation:
+        openAPIV3Schema:
+          properties:
+            service:
+              description: "The service that should have audit logging enabled. 
+                Example: 'storage.googleapis.com' for Cloud Storage.
+                'allServices' is a special value that covers all services"
+              type: string
+            log_types:
+              description: "The log types that should be enabled. The valudes are numerical.
```
joecheuk: Do you think we should create the mapping underneath and let user enter the actual log type in string instead? That way is less confusing for users.

t12g: Done. PTAL.

t12g:
The log types are now strings instead of ints. PTAL.

policies/templates/gcp_iam_audit_log.yaml
```diff
@@ -0,0 +1,98 @@
...
+spec:
+  crd:
+    spec:
+      names:
+        kind: GCPIAMAuditLogConstraintV1
+        plural: gcpiamauditlogconstraintsv1
+      validation:
+        openAPIV3Schema:
+          properties:
+            service:
```
joecheuk: if user wants to make sure DATA_READ & DATA_WRITE is enabled on say 5 apis, do we expect the user to create 5 constraints for that or should we allow them to specify that in one constraint? (i.e. having service as a list instead)

t12g: Done. We now accept a services list.

PR #78, opened 2019-06-03
move inline comment directive to be in the same line as the pipe

t12g:
So that we don't run into the following yaml parsing issue:
```
ERROR: logging before flag.Parse: I0603 15:52:13.667529  216768 validator.go:116] Unable to convert file policy-library/policies/templates/gcp_bq_dataset_location_v1.yaml, with error unmarshal []byte to yaml failed: yaml: line 86: did not find expected key, assuming this file should be skipped and continuing
```


t12g:
Yes, I tried it with terraform validator. There's no parsing error. To be sure, I also added always violate constraint, and it's giving out the expected error.

PR #80, opened 2019-06-05
Reorganize fixtures for dataset location and compute zone constraints

ddremund:
This PR is a simple reorganization of asset and constraint fixtures to conform to the structure discussed in #63.


morgante:
@ddremund I'm not sure about using the full constraint name for the directories. Instead, could we have it correspond to the rule filename? (ex. `cmek_rotation`)

ddremund:
@morgante Let's continue that conversation in #63 and settle on something.

ddremund:
@morgante Updated PR to conform to discussion in #63.

PR #91, opened 2019-06-07
Add a "Compute" prefix for the GCE VM external IP access template

t12g:


t12g:
@morgante done. Since we've renamed the template, I didn't rev the version. I only added the existing copy in legacy.

PR #101, opened 2019-06-12
Update constraint files

t12g:
Remove `gcp:` wrapper field for files under `/samples'.
Remove the matcher altogether for test fixture files.


joecheuk:
Does this break the old FCV instance?

t12g:
@joecheuk the changes under `/samples` will not work with old fcv instance. Is that a concern?

joecheuk:
Is it possible to make the FCV instance backward compatible? We have been advertising to users to sync their repos w/ the policy-libray repos and if they are already using some policies from the policy-library, updating to new FCV version will break all their existing constraints & pulling the changes down from the policy-library break all their current settings if they don't have their FCV updated. We should define the support policy so users can expect how many versions we support for backward compatibility vs releasing breaking changes without letting the users know.

t12g:
So the FCV instance is backwards compatible. Updating FCV will not break existing constraints, as it understands both formats (with and without the `gcp` wrapper field). 

What's incompatible in this PR is that an old FCV instance will ignore the matchers defined in the sample constraints. However, they're all "match all" anyways (the default behavior). Does that address your concern? Alternatively, we could also remove the `match` block everywhere. WDYT?

joecheuk:
Got it, thanks for the clarifications!

PR #102, opened 2019-06-17
Adding constraint label_enforce

AdrienWalkowiak:
This constraint checks for a list of labels to be present on specific resource types that do support labels. The support types so far are:

* cloudresourcemanager.googleapis.com/Project
* storage.googleapis.com/Bucket
* compute.googleapis.com/Instance
* compute.googleapis.com/Image
* compute.googleapis.com/Disk
* compute.googleapis.com/Snapshot

The future steps for this constraint will be to:
1. Let the use specify which types to scan for as a parameter
2. Add/validate/test more resource types 
2. Add a way to check the format of each label's value (using a regexp or similar)


validator/enforce_labels_test.rego
```diff
@@ -0,0 +1,201 @@
...
+test_enforce_label_projects_violates_project_four {
+	violations := project_violations
+	count(violations) == 4
+	violation := violations[_]
+	is_string(violation.msg)
+	is_object(violation.details)
+}
+
+# confirm which 3 projects are in violation
+test_enforce_label_projects_violates_project_basic {
```
morgante: If we're looking for 3 projects, why do we count 4 above?

validator/enforce_labels_test.rego
```diff
@@ -0,0 +1,201 @@
...
+test_enforce_label_compute_instance_violates_project_four {
+	violations := compute_instance_violations
+	count(violations) == 4
+	violation := violations[_]
+	is_string(violation.msg)
+	is_object(violation.details)
+}
+
+# confirm which 3 instances are in violation
+test_enforce_label_compute_instance_violates_project_basic {
```
morgante: If we're looking for 3, why do we count 4 above?

validator/enforce_labels_test.rego
```diff
@@ -0,0 +1,201 @@
...
+test_enforce_label_compute_instance_violates_project_four {
+	violations := compute_image_violations
+	count(violations) == 4
+	violation := violations[_]
+	is_string(violation.msg)
+	is_object(violation.details)
+}
+
+# confirm which 3 images are in violation
+test_enforce_label_compute_image_violates_project_basic {
```
morgante: If we're looking for 3, why do we count 4 above?

validator/enforce_labels_test.rego
```diff
@@ -0,0 +1,201 @@
...
+test_enforce_label_compute_disk_violates_project_four {
+	violations := compute_disk_violations
+	count(violations) == 4
+	violation := violations[_]
+	is_string(violation.msg)
+	is_object(violation.details)
+}
+
+# confirm which 3 disks are in violation
+test_enforce_label_compute_disk_violates_project_basic {
```
morgante: If we're looking for 3, why do we count 4 above?

validator/enforce_labels_test.rego
```diff
@@ -0,0 +1,201 @@
...
+test_enforce_label_compute_snapshot_violates_project_four {
+	violations := compute_snapshot_violations
+	count(violations) == 4
+	violation := violations[_]
+	is_string(violation.msg)
+	is_object(violation.details)
+}
+
+# confirm which 3 snapshots are in violation
+test_enforce_label_compute_snapshot_violates_project_basic {
```
morgante: If we're looking for 3, why do we count 4 above?

validator/enforce_labels_test.rego
```diff
@@ -0,0 +1,201 @@
...
+	violations := compute_disk_violations
+
+	violations[_].details.resource == "//compute.googleapis.com/projects/my-test-project/zones/us-east1-b/disks/instance-1-invalid-disk-missing-labels"
+	violations[_].details.resource == "//compute.googleapis.com/projects/my-test-project/zones/us-east1-b/disks/instance-1-invalid-disk-missing-label1"
+	violations[_].details.resource == "//compute.googleapis.com/projects/my-test-project/zones/us-east1-b/disks/instance-1-invalid-disk-missing-label2"
+}
+
+#### Testing for GCE Snapshots
+# # Confirm exactly 4 snapshot violations were found
+test_enforce_label_compute_snapshot_violates_project_four {
```
morgante: This name seems off. Why do we include `project` in all of them?

```suggestion
test_enforce_label_compute_snapshot_violates_project_four {
```

validator/enforce_labels.rego
```diff
@@ -0,0 +1,69 @@
...
+	mandatory_labels := params.mandatory_labels[_]
+	resource_is_to_scan(asset)
+	not label_is_valid(mandatory_labels, asset)
+
+	message := sprintf("%v doesn't have a required label.", [asset.name])
+
+	metadata := {"resource": asset.name, "mandatory_labels": mandatory_labels}
+}
+
+resource_is_to_scan(asset) {
```
morgante: I think a best practice is to have an explicit return value for the function.

```suggestion
resource_is_to_scan(asset) = true {
```

validator/enforce_labels.rego
```diff
@@ -0,0 +1,69 @@
...
+	resource_types_to_scan[_] == asset.asset_type
+}
+
+# generic label_is_valid for all resources
+label_is_valid(label, asset) {
```
morgante:
```suggestion
label_is_valid(label, asset) = true {
```

policies/templates/gcp_enforce_labels_v1.yaml
```diff
@@ -0,0 +1,100 @@
...
+  name: gcp-enforce_label-v1
+spec:
+  crd:
+    spec:
+      names:
+        kind: GCPEnforceLabelConstraintV1
+        plural: gcpenforcelabelconstraintsv1
+      validation:
+        openAPIV3Schema:
+          properties: {}
```
morgante: We should include `mandatory_labels` property here.

morgante:
> Add a way to check the format of each label's value (using a regexp or similar)

For this part, it's worth taking a look at [this PR](https://github.com/forseti-security/policy-library/pull/76).

validator/enforce_labels_test.rego
AdrienWalkowiak: we have 3 projects with violations, one of which has 2 violations (one violation per missing label).

so the first test checks that we do have 4 violations, the second one test that the right 3 bad projects are the ones flagged. This is the same for all the other tests.

validator/enforce_labels_test.rego
AdrienWalkowiak: true, the name is wrong, will fix

validator/enforce_labels_test.rego
morgante: Got it, might help to have a clarifying comment.

AdrienWalkowiak:
Applied the recommended changes:
- added the properties for the template 
- Added some clarifying comments in the tests
- Renamed test functions to remove the _project_  typos
- Added explicit return value in the rule for 2 functions

morgante:
> Added explicit return value in the rule for 2 functions

I don't see this change. Also, make sure to run `make build` once you do.

AdrienWalkowiak:
It looks like make format (and thus make build) remove the = true from the function definition... What do you think I should do?

morgante:
In that case, let's leave it as is.

validator/enforce_labels.rego
morgante: Turns out `opa fmt` removes this automatically.

PR #105, opened 2019-06-18
[WIP] Add a simple storage class constraint

morgante:


AdrienWalkowiak:
@morgante is this sill WIP?

morgante:
I don't think this constraint is practically very useful, so closing it.

PR #108, opened 2019-06-20
enforce_label update: Adding new features + new resource types

AdrienWalkowiak:
New features:

- Adding optional parameter to provide a list of resource types to scan for.
- Adding feature to specify the pattern mandatory label's values should match (list of  label: pattern object in parameters)

Support for new resource types:

- Cloud SQL instances
- BigTable instances

### DIFF
--- a/policies/templates/gcp_enforce_labels_v1.yaml
+++ b/policies/templates/gcp_enforce_labels_v1.yaml
@@ -110,7 +110,8 @@ spec:
        
        	resource_types_to_scan := lib.get_default(params, "resource_types_to_scan", tested_resource_types)
        
-       	resource_is_to_scan(asset, resource_types_to_scan)
+       	# test if resource needs to be scanned
+       	resource_types_to_scan[_] == asset.asset_type
        
        	not label_is_valid(label_key, label_value_pattern, asset, standard_types)
        
@@ -118,11 +119,7 @@ spec:
        	metadata := {"resource": asset.name, "label_in_violation": label_key}
        }
        
-       resource_is_to_scan(asset, resource_types_to_scan) {
-       	resource_types_to_scan[_] == asset.asset_type
-       }
-       
-       # check if label exists for all resources to scan
+       # check if label exists and if its value matches the pattern passed as a parameter for all resources to scan
        label_is_valid(label_key, label_value_pattern, asset, standard_types) {
        	# retrieve the right values from asset
        	resource_labels := get_labels(asset, standard_types)
--- a/policies/templates/gcp_enforce_labels_v1.yaml
+++ b/policies/templates/gcp_enforce_labels_v1.yaml
@@ -99,13 +99,21 @@ spec:
        		"sqladmin.googleapis.com/Instance",
        	]
        
-       	non_standard_types := ["sqladmin.googleapis.com/Instance"]
+       	standard_types := [
+       		"cloudresourcemanager.googleapis.com/Project",
+       		"storage.googleapis.com/Bucket",
+       		"compute.googleapis.com/Instance",
+       		"compute.googleapis.com/Image",
+       		"compute.googleapis.com/Disk",
+       		"compute.googleapis.com/Snapshot",
+       		"google.bigtable.Instance",
+       	]
        
        	resource_types_to_scan := lib.get_default(params, "resource_types_to_scan", tested_resource_types)
        
        	resource_is_to_scan(asset, resource_types_to_scan)
        
-       	not label_is_valid(label_key, label_value_pattern, asset, non_standard_types)
+       	not label_is_valid(label_key, label_value_pattern, asset, standard_types)
        
        	message := sprintf("%v's label is in violation.", [asset.name])
        	metadata := {"resource": asset.name, "label_in_violation": label_key}
@@ -116,9 +124,9 @@ spec:
        }
        
        # check if label exists for all resources to scan
-       label_is_valid(label_key, label_value_pattern, asset, non_standard_types) {
+       label_is_valid(label_key, label_value_pattern, asset, standard_types) {
        	# retrieve the right values from asset
-       	resource_labels := get_labels(asset, non_standard_types)
+       	resource_labels := get_labels(asset, standard_types)
        
        	# test if label exists in asset
        	resource_labels[label_key]
@@ -128,14 +136,14 @@ spec:
        }
        
        # get_labels for standard resources
-       get_labels(asset, non_standard_types) = resource_labels {
-       	asset.asset_type != non_standard_types[_]
+       get_labels(asset, standard_types) = resource_labels {
+       	asset.asset_type == standard_types[_]
        	resource := asset.resource.data
        	resource_labels := lib.get_default(resource, "labels", {})
        }
        
        # get_labels for cloudsql instances
-       get_labels(asset, non_standard_types) = resource_labels {
+       get_labels(asset, standard_types) = resource_labels {
        	asset.asset_type == "sqladmin.googleapis.com/Instance"
        	resource := asset.resource.data.settings
        	resource_labels := lib.get_default(resource, "userLabels", {})
--- a/policies/templates/gcp_enforce_labels_v1.yaml
+++ b/policies/templates/gcp_enforce_labels_v1.yaml
@@ -35,13 +35,25 @@ spec:
                 "
               type: array
               items: objects
+              required: true
             resource_types_to_scan:
               description: | 
                 "Optional parameter: list of resource types to scan for labels.
                 Any resource that is not of these types will not raise any violation.
-                if not passed, all tested resource types will be scanned for violations (see below for full list)"
+                if not passed, all tested resource types will be scanned for violations:
+
+                - cloudresourcemanager.googleapis.com/Project
+                - storage.googleapis.com/Bucket
+                - compute.googleapis.com/Instance
+                - compute.googleapis.com/Image
+                - compute.googleapis.com/Disk
+                - compute.googleapis.com/Snapshot
+                - google.bigtable.Instance
+                - sqladmin.googleapis.com/Instance
+                "
               type: array
               items: string
+              required: false
   targets:
     validation.gcp.forsetisecurity.org:
       rego: | #INLINE("validator/enforce_labels.rego")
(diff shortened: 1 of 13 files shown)


policies/templates/gcp_enforce_labels_v1.yaml
```diff
@@ -26,79 +26,106 @@ spec:
...
+                "List of mandatory label objects to check for on all supported resources. 
+                Label objects should be as follow:
+                label_key: label_value_regex_to_match 
+                Each missing label will result in a violation for scanned resources.
+                Each mismatching label will result in a violation for scanned resources.
+                "
+              type: array
+              items: objects
+            resource_types_to_scan:
+              description: |
```
t12g: instead of specifying whether something is optional in the comment, have we considered using the `required` property to denote the required fields? https://tools.ietf.org/html/draft-wright-json-schema-validation-00#section-5.15

validator/enforce_labels.rego
```diff
@@ -35,35 +26,58 @@ deny[{
...
-	message := sprintf("%v doesn't have a required label.", [asset.name])
+	resource_types_to_scan := lib.get_default(params, "resource_types_to_scan", tested_resource_types)
 
-	metadata := {"resource": asset.name, "mandatory_labels": mandatory_labels}
+	resource_is_to_scan(asset, resource_types_to_scan)
+
+	not label_is_valid(label_key, label_value_pattern, asset, non_standard_types)
+
+	message := sprintf("%v's label is in violation.", [asset.name])
+	metadata := {"resource": asset.name, "label_in_violation": label_key}
```
t12g: does this mean we'll have one violation per missing/invalid label? do we want to generate one violation per asset regardless of how many labels are missing/invalid instead?

validator/enforce_labels.rego
```diff
@@ -35,35 +26,58 @@ deny[{
...
+
+	# test if label value matches pattern passed as a parameter 
+re_match(	label_value_pattern, resource_labels[label_key])
 }
 
-# get_generic_label Object for standard resources
-get_generic_label(asset) = resource {
+# get_labels for standard resources
+get_labels(asset, non_standard_types) = resource_labels {
+	asset.asset_type != non_standard_types[_]
```
t12g: is this right? we seem to be checking that asset_type != **any** one of the standard types here.

validator/enforce_labels.rego
```diff
@@ -35,35 +26,58 @@ deny[{
...
+	resource_labels := lib.get_default(resource, "labels", {})
 }
 
-# get_label for standard resources
-# placeholder function: in case we run into non-standard location for labels
-
-get_label(asset) = resource {
-	resource_types_to_scan[_] == asset.asset_type
-	resource := get_generic_label(asset)
+# ge_labels for cloudsql instances
```
t12g: sp: get

validator/enforce_labels.rego
AdrienWalkowiak: You will indeed get one violation per missing/invalid label, which was the intent here. This is easier to troubleshoot for the end user, and in the context of multiple people working on the violations (for instance different teams working on the same assets), this makes it easier to split the tasks / fixes.

policies/templates/gcp_enforce_labels_v1.yaml
```diff
@@ -26,79 +26,106 @@ spec:
...
+                Each missing label will result in a violation for scanned resources.
+                Each mismatching label will result in a violation for scanned resources.
+                "
+              type: array
+              items: objects
+            resource_types_to_scan:
+              description: | 
+                "Optional parameter: list of resource types to scan for labels.
+                Any resource that is not of these types will not raise any violation.
+                if not passed, all tested resource types will be scanned for violations (see below for full list)"
```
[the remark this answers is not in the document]
AdrienWalkowiak: Sure, I will add it to the next commit.

validator/enforce_labels.rego
AdrienWalkowiak: I think it's correct, this test will return false if the asset type is non-standard which is what we want here: only standard types have the labels under "labels" in asset.resource.data.

Each non-standard type (or group of non-standard types) will have their own get_labels function to retrieve the labels in their custom location (like the sql instances in this version).

validator/enforce_labels.rego
AdrienWalkowiak: will fix in next commit

policies/templates/gcp_enforce_labels_v1.yaml
AdrienWalkowiak: I think that would help clarify the usage here, if it's possible. I have not seen it used in the other constraints, but they may not have optional parameters...

AdrienWalkowiak: I am not sure how to implement that the right way, is it a simple

`required: true` as part of mandatory_labels (and the opposite for the optional parameter)?

t12g: My understanding is that `required` field take a list of field names; it's at the `properties` level. 
So it'd look something like
```
properties:
  mandatory_labels:
   ...
required: ["mandatory_labels"]
```

Also while we're here, should we also define the schema for the objects in `mandatory_labels`

validator/enforce_labels.rego
t12g: oops. I meant to say checking that asset_type != any one of the non-standard types. It happens to work here because there's only 1 non-standard type. But if there's more than 1, it's not going to give the correct result. Example: https://play.openpolicyagent.org/p/AY0g0knZ2r

AdrienWalkowiak: Good catch! I did not think about this. I flipped the logic around to test exactly for standard types instead since this is what I was testing for anyway, it seems to work fine.

policies/templates/gcp_enforce_labels_v1.yaml
AdrienWalkowiak: I updated the required property as you mentioned, that makes sense. 

I am not sure how to define the objects in mandatory_labels. Do you have an example I can reuse?

t12g: unfortunately I don't think we have an existing example in policy-library, but I found one example on stackoverflow: https://stackoverflow.com/questions/49822794/kubernetes-custom-resource-definition-required-field

policies/templates/gcp_enforce_labels_v1.yaml
```diff
@@ -26,79 +26,125 @@ spec:
...
+       	]
+       
+       	standard_types := [
+       		"cloudresourcemanager.googleapis.com/Project",
+       		"storage.googleapis.com/Bucket",
+       		"compute.googleapis.com/Instance",
+       		"compute.googleapis.com/Image",
+       		"compute.googleapis.com/Disk",
+       		"compute.googleapis.com/Snapshot",
+       		"google.bigtable.Instance",
```
t12g: this type doesn't seem to match https://cloud.google.com/resource-manager/docs/cloud-asset-inventory/overview#
is this from a pre-v1 CAI API invocation?

policies/templates/gcp_enforce_labels_v1.yaml
```diff
@@ -26,79 +26,125 @@ spec:
...
+       	]
+       
+       	resource_types_to_scan := lib.get_default(params, "resource_types_to_scan", tested_resource_types)
+       
+       	resource_is_to_scan(asset, resource_types_to_scan)
```
t12g: nit: is this abstraction useful? Seems that inlining `resource_types_to_scan[_] == asset.asset_type` is easier to follow

policies/templates/gcp_enforce_labels_v1.yaml
```diff
@@ -26,79 +26,125 @@ spec:
...
+       
+       	message := sprintf("%v's label is in violation.", [asset.name])
+       	metadata := {"resource": asset.name, "label_in_violation": label_key}
+       }
+       
+       resource_is_to_scan(asset, resource_types_to_scan) {
+       	resource_types_to_scan[_] == asset.asset_type
+       }
+       
+       # check if label exists for all resources to scan
```
t12g: nit: this comment isn't exactly accurate, since we're checking for label value in addition to existence

policies/templates/gcp_enforce_labels_v1.yaml
AdrienWalkowiak: Indeed this type is different, not too sure why, it does comes from a v1 CAI API export.

policies/templates/gcp_enforce_labels_v1.yaml
AdrienWalkowiak: true, I wil lfix it in the next commit.

policies/templates/gcp_enforce_labels_v1.yaml
AdrienWalkowiak: true, I forgot to update this one. Wil lfix in next commit

policies/templates/gcp_enforce_labels_v1.yaml
AdrienWalkowiak: This example helps, but I am still not sure of the exact syntax for this use case, since there is not a fixed name for the objects in that list, its pretty much a list of objects like:

`label_key: label_value_pattern
`

so it's a dynamic field... I wil do some research on this.

validator/enforce_labels_test.rego
```diff
@@ -86,122 +88,185 @@ compute_snapshot_violations[violation] {
...
 	violations := project_violations
-	count(violations) == 4
+	count(violations) == 6
 	violation := violations[_]
 	is_string(violation.msg)
 	is_object(violation.details)
 }
 
-# confirm which 3 projects are in violation
+# confirm which 4 projects are in violation
```
t12g: can we also verify that there are no other projects in violation (similarly below)? otherwise we're only checking for subset, right?

example: https://github.com/forseti-security/policy-library/blob/master/validator/gke_allowed_node_sa_scopes_test.rego#L51

AdrienWalkowiak: Great thinking. I updated all the tests to reflect this change and I merged the count and comparison ones into a single test, so that it's easier to read and maintain.

PR #113, opened 2019-07-06
Update user_guide.md

kojnp:
added mention that current validator doesn't work with terraform 0.12

### DIFF
--- a/docs/user_guide.md
+++ b/docs/user_guide.md
@@ -250,7 +250,8 @@ chmod 755 terraform-validator-linux-amd64
 
 ### For local development environments
 
-These instructions assume you have forked a branch and is working locally. Also it assumes that you have terraform v0.11 . Currently using terraform v0.12 is not supported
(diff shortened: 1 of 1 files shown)


kojnp:
CLA: I signed it!

docs/user_guide.md
```diff
@@ -250,7 +250,7 @@ chmod 755 terraform-validator-linux-amd64
 
 ### For local development environments
 
-These instructions assume you have forked a branch and is working locally.
+These instructions assume you have forked a branch and is working locally. Also it assumes that you have terraform v0.11 . Currently using terraform v0.12 is not supported
```
briantkennedy: Thanks for the fix.  One comment, these two sentences effectively repeat the same idea, could you please rephrase as "Only terraform v0.11 is supported".

kojnp: rephrased - pls review

PR #116, opened 2019-07-15
Fix Iam binding issue + add new supported types

AdrienWalkowiak:
Adding fix for IAM Binding false positives in Forseti. 
Defaulting to the standard location for untested resource types (resource.data.labels)
Adding tests/support for dataproc clusters and jobs


AdrienWalkowiak:
Actually in Forseti you have false positive because of the IAM binding resources that get created with projects or buckets for instance. The data object is empty for these resources so I added a line in the rule to test that asset.resource.data is not empty to fix this.

policies/templates/gcp_enforce_labels_v1.yaml
```diff
@@ -84,10 +90,13 @@ spec:
        	lib.get_constraint_params(constraint, params)
        	asset := input.asset
        
+       	# test if we have a valid resource
+       	resource := asset.resource.data
```
AdrienWalkowiak: @morgante that's the additional test I was mentioning to avoid false positives

PR #122, opened 2019-07-19
Removed non-supported resources

jdyke:
While going through some validations of pre-existing templates and constraints, I noticed unsupported asset_types / terraform resources were used as examples. This can lead to end user confusion so I removed from the .rego, template, and example constraints file. I left the code inside the test file but that can be easily removed as well.


morgante:
Just because Terraform Validator doesn't support these doesn't mean we should remove them. In particular, Forseti can still validate these resource types.

Issue #124, opened 2019-07-23
sqladmin.googleapis.com/Instance requireSSL Not Triggering


jdyke:
gcp_sql_ssl_v1.yaml template:

Is looking for:

`asset.resource.settings.ipConfiguration.requireSsl == false`

but the CAI returns:

```
[
  {
    "name": "<redacted>",
    "asset_type": "sqladmin.googleapis.com/Instance",
    "ancestry_path": "<redacted>",
    "resource": {
      "version": "v1beta4",
      "discovery_document_uri": "https://www.googleapis.com/discovery/v1/apis/sqladmin/v1beta4/rest",
      "discovery_name": "DatabaseInstance",
      "parent": "<redacted>",
      "data": {
        "databaseVersion": "POSTGRES_9_6",
        "name": "master-instance",
        "project": "<redacted>",
        "region": "us-central1",
        "settings": {
          "ipConfiguration": {
            "ipv4Enabled": true,
            "requireSsl": false
          },
          "pricingPlan": "PER_USE",
          "replicationType": "SYNCHRONOUS",
          "storageAutoResize": true,
          "tier": "db-f1-micro"
        }
      }
    }
  }
]
```

 It looks like it is missing "data" so I added it to the template: 

`asset.resource.data.settings.ipConfiguration.requireSsl == false`

but that still does not catch the entry:

```
variable project {}

provider "google" {
  project = "${var.project}"
  region  = "us-east1"
}

resource "google_sql_database_instance" "master" {
  name = "master-instance"
  database_version = "POSTGRES_9_6"
  region = "us-central1"

  settings {
    # Second-generation instance tiers are based on the machine
    # type. See argument reference below.
    tier = "db-f1-micro"
    ip_configuration {
        ipv4_enabled = "true"
        require_ssl = "false"
    }
  }
}
```

morgante:
Please test in Forseti.

morgante:
@jdyke I opened #125 to address the issue in the constraint template.

I tested the conversion in Terraform Validator and it looks like it properly converts the requireSsl prop:

```
{
	"name": "//cloudsql.googleapis.com/projects/gcp-foundation-shared-devops/instances/master-instance",
	"asset_type": "sqladmin.googleapis.com/Instance",
	"ancestry_path": "organization/816421441114/project/gcp-foundation-shared-devops",
	"resource": {
		"version": "v1beta4",
[...]

PR #126, opened 2019-07-23
Fix/sql allowed authorized networks.rego

FanchenBao:
The rego file "sql_allowed_authorized_networks.rego" has two issues.
1. In the package statement, `templates.gcp.GCPSQLAllowedAuthorizedNetworksV1` needs be changed into `templates.gcp.GCPSQLAllowedAuthorizedNetworksConstraintV1` in order for the template file "gcp_sql_allowed_authorized_networks_v1.yaml" to work.
2. `data` field needs to be added in between `resource` and `settings` to correctly retrieve info from `authorizedNetworks`.

I also modified the logic so that the rule can detect when `authorizedNetworks` does not exist or has no content.


FanchenBao:
I signed CLA.

FanchenBao:
#124

FanchenBao:
You are welcome. And thanks for the suggestion. I will work on it.

FanchenBao:
test feature and test rego file for sql_allowed_authorized_networks have been updated. All tests passed using `make test`.

AdrienWalkowiak:
@morgante any other comments on this?

policies/templates/gcp_sql_allowed_authorized_networks_v1.yaml
```diff
@@ -71,9 +71,14 @@ spec:
             	asset.asset_type == "sqladmin.googleapis.com/Instance"
             
             	allowed_authorized_networks = lib.get_default(params, "authorized_networks", [])
-            	configured_networks := {network |
-            		network = asset.resource.settings.ipConfiguration.authorizedNetworks[_].value
-            	}
+            	
+                # Check whether authorizedNetworks field exists, so that
+                # we can report violation when this field is not set
+                config_auth_networks = lib.get_default(asset.resource.data.settings.ipConfiguration, "authorizedNetworks", [{"value":"authorized network unspecified"}])
```
morgante: Formatting looks off. Please run `make format`.

policies/templates/gcp_sql_allowed_authorized_networks_v1.yaml
```diff
@@ -71,9 +71,14 @@ spec:
...
+
+                configured_networks := {network |
+                network = config_auth_networks[_].value
+                }
             
             	matched_networks := {network |
```
morgante: Instead of using a comprehension here, it should be sufficient to cast `allowed_authorized_networks` to a set then compare the two sets (`forbidden_networks := configured_networks - allowed_authorized_networks`).

FanchenBao: `allowed_authorized_networks` is the full whitelist, which I think is not what we want to compare here. We want to compare `configured_networks` with the networks matched to `allowed_authorized_networks`. That's why the comparison has to be between `configured_networks` and `matched_networks`.

As for `configured_networks`, I think it has to be a comprehension because we want to extract each value from `config_auth_networks`, which itself is an array of key/value maps.

policies/templates/gcp_sql_allowed_authorized_networks_v1.yaml
FanchenBao: File is now formatted.

ocervell:
@AdrienWalkowiak @FanchenBao we shouldn't edit the fixture `data.json` manually - IMHO it should come from a real environment, otherwise we end up with issues where CAI data is different than our test fixtures

FanchenBao:
If I remember correctly, the real environment CAI has `data` field, at least when I made the change last month. Therefore, I changed `data.json` file to reflect that fact.

morgante:
@FanchenBao Can you rebase?

FanchenBao:
@morgante Just rebased my fork onto current master.

morgante:
@FanchenBao You'll have to resolve the conflicts manually. Also probably worth looking at the current implementation on master and seeing if your change still makes sense.

morgante:
@FanchenBao Looks like tests fail, likely due to the addition of `data` in the test data.

Can you please run a CAI export and verify that it actually has that structure? Then make sure `make test` passes.

FanchenBao:
@morgante Yes, at least from my end, the CAI file has "data" field. For example:
```
[ 
   { 
      "name":"//cloudsql.googleapis.com/projects/fanchen-sandbox/instances/private-instance",
      "asset_type":"sqladmin.googleapis.com/Instance",
      "ancestry_path":"project/fanchen-sandbox",
      "resource":{ 
         "version":"v1beta4",
         "discovery_document_uri":"https://www.googleapis.com/discovery/v1/apis/sqladmin/v1beta4/rest",
         "discovery_name":"DatabaseInstance",
         "parent":"//cloudresourcemanager.googleapis.com/projects/fanchen-sandbox",
         "data":{ 
            "databaseVersion":"MYSQL_5_7",
            "name":"private-instance",
            "project":"fanchen-sandbox",
            "region":"us-central1",
            "settings":{ 
               "ipConfiguration":{ 
                  "authorizedNetworks":[ 
                     { 
                        "name":"test_2",
                        "value":"199.27.25.0/24"
                     }
                  ],
                  "ipv4Enabled":false,
                  "requireSsl":true
               }
            }
         }
      }
   }
]
```
I also double-checked `data.json` files in other sql-related assets, and they all have "data" field. Furthermore, I found that my old logic of reporting violation when `authorizedNetworks` is not set is incorrect, because this field is optional. Therefore, I adopted the current code on master branch.

PR #128, opened 2019-07-24
Fixed CAI API plurality issue for networkInterfaces

### LINKED ISSUE #10: External IP constraint doesn't detect ephemeral IPs
Currently, if you create a new VM with an external IP via Terraform with an ephemeral IP the constraint doesn't violate.

onetwopunch:
The [CAI API object for compute instances](https://www.googleapis.com/discovery/v1/apis/compute/v1/rest) has the block `networkInterfaces`
and within that `accessConfigs`, but the validator for `vm_external_ip` as well as
multiple fixtures, use the singular version. Another issue with the `vm_external_ip`
validator was that at the time of the terraform plan, an `external_ip` will not
necessarily exist, so checking for the existence of an `access_config` block is a more
stable solution.

Fixes #10


morgante:
@onetwopunch Unfortunately the CAI data actually exposes the singular, while the API itself (and Terraform Validator) uses the plural versions. We are waiting on an internal bug to resolve this (towards using the plural).

policies/templates/gcp_compute_external_ip_access_v1.yaml
```diff
@@ -71,8 +71,7 @@ spec:
             
             	# Find network access config block w/ external IP
             	instance := asset.resource.data
-            	access_config := instance.networkInterface[_].accessConfig
-            	external_ip := access_config[_].externalIp
+            	access_config := instance.networkInterfaces[_].accessConfigs
```
morgante: We should also check that at least one access config is included. If the `accessConfigs` array is present but has no members, there isn't an external IP.

onetwopunch: Good call will fix today

morgante:
I just need to check the output of new CAI dumps.

Issue #115, opened 2019-07-15
Template for whitelisting selected IAM roles


marcin-kolda:
We would like to have a template which allows only selected roles for given users. All other IAM bindings should be reported as violations. So instead of blacklisting all other roles, we would like to whitelist only few of them.

Example:
```yaml
apiVersion: constraints.gatekeeper.sh/v1alpha1
kind: GCPIAMAllowedRolesConstraintV1
metadata:
  name: whitelist_roles
  annotations:
    description: Only test.com groups can have appengine.appViewer and logging.viewer roles. All others bindings are not allowed.
spec:
  severity: high
  match:
    target: ["organization/*"]
    exclude: [] # optional, default is no exclusions
  parameters:
    role:
    - roles/appengine.appViewer
    - roles/logging.viewer
    members:
    - "group:*@test.com"
```

morgante:
Have you looked at the https://github.com/forseti-security/policy-library/blob/master/samples/iam_restrict_role.yaml constraint? It seems like it should solve this use case.

marcin-kolda:
Hi @morgante, unfortunately iam_restrict_role.yaml doesn't work for me (see #117). As I understand it validates the minimum required roles (e.g. all project require project viewer role) in whitelisting mode.
In my use case I'd like to define possible roles which can be used in IAMs for given members, e.g. few read roles for users, deploy roles for service accounts. All other IAM bindings should be reported as violations.

marcin-kolda:
After clarification in #117 the iam_restrict_role.yaml rule still will not work in my case.

I would like to whitelist specific roles and blacklist everything else. I don't want to specify all possible roles for blacklisting, as new one are added dynamically. I'd like to define few roles as whitelisted, which were evaluated and approved. Everything else should be marked as violation.

To give you more context, we want to use Terraform Validator to validate other teams IAM changes in our CI pipeline. So we might end up using different rules for different access on the same project, e.g. service accounts for deploy, user account for read only.

We could also use combination of 2 rules:
- multiple existing iam_restrict_role.yaml rules in whitelisting mode for every approved role,
- new rule, which would use the same allowed roles and blacklist every other role

morgante:
Thanks for clarifying, I would suggest solving this through a combination of a few constraint:
1. A whitelist (using `iam_restrict_role`) of approved members for those each approved role
2. A new constraint template to whitelist only the approved roles (and block any other role)

[...]

PR #130, opened 2019-07-26
Add/rule for allow ban roles

FanchenBao:
Add new template (`gcp_iam_allow_ban_roles_v1`) and constraints (`iam_allowed_roles`, `iam_banned_roles`) to allow or ban target roles, inspired by #115.

Differences between `gcp_iam_allow_ban_roles_v1` and `gcp_iam_allowed_bindings_v1` are:
 1. `gcp_iam_allowed_bindings_v1` cannot whitelist roles (it can whitelist members assigned to a defined role, but not the role itself). `gcp_iam_allow_ban_roles_v1` filled in this gap.
 2. `gcp_iam_allowed_bindings_v1` can only blacklist one role at a time (i.e. n constraints are needed to blacklist n roles ). `gcp_iam_allow_ban_roles_v1` can ban multiple roles in one constraint.

"allow" and "ban" are chosen for the new template and constraints in order not to cause confusion with the existing terms of "whitelist" and "blacklist" associated with `gcp_iam_allowed_bindings_v1` and its sample constraints.

Test cases and test rego code are added as well. All tests pass.


AdrienWalkowiak:
@joecheuk @morgante @t12g @briantkennedy can any of you look at this one? It looks relevant to me...

validator/iam_allow_ban_roles.rego
```diff
@@ -0,0 +1,59 @@
...
+	asset := input.asset
+
+	binding := asset.iam_policy.bindings[_]
+	role := binding.role
+
+	matches_found = {r | r := role; glob.match(params.roles[_], [], r)}
+
+	mode := lib.get_default(params, "mode", "allow")
+
+	target_match_count(mode, desired_count)
```
briantkennedy: I'm a little new to rego, so I'm not sure what this is doing here.  [The rego docs](https://www.openpolicyagent.org/docs/latest/how-do-i-write-policies/#functions) don't seem to define what happens if a function is given more arguments than it's defined with, so it's not clear to me what the value of "desired_count" will evaluate to here.  What does this value represent?

FanchenBao: I copied this implementation from "iam_allowed_bindings.rego". I think `desired_count` stores the returned value of `target_match_count`. If `mode == "ban"`, `desired_count` evaluates to 0, otherwise 1.

morgante: This code is equivalent:

```
desired_count := target_match_count(mode)
```

validator/iam_allow_ban_roles.rego
```diff
@@ -0,0 +1,59 @@
...
+	count(matches_found) != desired_count
+
+	message := sprintf("IAM policy for %v grants %v", [asset.name, role])
```
briantkennedy: Add note about whether the rule was a deny or allow

validator/iam_allow_ban_roles.rego
```diff
@@ -0,0 +1,59 @@
...
+
+	binding := asset.iam_policy.bindings[_]
+	role := binding.role
+
+	matches_found = {r | r := role; glob.match(params.roles[_], [], r)}
+
+	mode := lib.get_default(params, "mode", "allow")
+
+	target_match_count(mode, desired_count)
+	count(matches_found) != desired_count
```
briantkennedy: If the user has two matches in parameters.roles (eg,["roles/viewer", "roles/*"]) this will fail on legal inputs for the allow case.  I think you need something like (match == "allow") == (0 < count(matches_found)).

FanchenBao: This situation would not happen, because these two lines
```
binding := asset.iam_policy.bindings[_]
role := binding.role
```
determines that we are checking the roles one by one. See [here](https://www.openpolicyagent.org/docs/v0.10.7/how-does-opa-work/) for the following quote

> Expressions can reference elements in a collection using the [_] and [<variable>] syntax. OPA knows to evaluate such queries by iterating over each element in the corresponding collection.

So in `matches_found = {r | r := role; glob.match(params.roles[_], [], r)}`, there is always only one value for `r` in each iteration. Therefore, value of `count(matches_found)` can only be 1 or 0.

briantkennedy: oh, my mistake, I misread the {r | r ...} as [r | r ...]

validator/iam_allow_ban_roles.rego
FanchenBao: Error message updated. The message now clearly states whether the violation is due to role being banned or not allowed.

validator/iam_allow_ban_roles.rego
FanchenBao: Code updated to use syntax `desired_count := target_match_count(mode)` as it is more readable than the original version.

PR #140, opened 2019-08-05
Disable default service account on GKE

charliewolf:


validator/gke_disable_default_service_account_test.rego
```diff
@@ -0,0 +1,40 @@
+#
+# Copyright 2018 Google LLC
```
AdrienWalkowiak: nit: update to 2019

validator/gke_disable_default_service_account_test.rego
```diff
@@ -0,0 +1,40 @@
...
+	resource := data.test.fixtures.gke_disable_default_service_account.assets[_]
+	constraint := data.test.fixtures.gke_disable_default_service_account.constraints.disable_gke_default_service_account
+
+	issues := deny with input.asset as resource
+		 with input.constraint as constraint
+
+	violation := issues[_]
+}
+
+test_nodepool_limited_service_account {
```
AdrienWalkowiak: Could we scale a bit these tests? I would like to check that the assets in violation are exactly equal to a set of asset names that should violate in our data set.
You can check [this pr](https://github.com/forseti-security/policy-library/blob/master/validator/allowed_resource_types_test.rego#L59) for instance

AdrienWalkowiak:
It looks like there are some issues with the latest changes:


`data.templates.gcp.GCPGKEDisableDefaultServiceAccountConstraintV1.test_nodepool_default_service_account: FAIL (4.09066ms)`

charliewolf:
@AdrienWalkowiak should be fixed now

AdrienWalkowiak:
Sorry to bug you again, but could you run make build too? The formatting will become an issue for future PRs as we automated tests on the use of opa fmt.

charliewolf:
@AdrienWalkowiak no worries, just did that in latest push

policies/templates/gcp_iam_allow_ban_roles_v1.yaml
```diff
@@ -107,3 +107,4 @@ spec:
             }
             
             #ENDINLINE
+            #ENDINLINE
```
AdrienWalkowiak: I don't think this was an intended change in this PR - could you revert it?

charliewolf:
@AdrienWalkowiak done

PR #143, opened 2019-08-05
Dataproc location dev

AdrienWalkowiak:
Adding Dataproc location policy


AdrienWalkowiak:
it's the same issue than in #145 - make format is breaking this format for re_match among other things.

AdrienWalkowiak:
I removed the default for the locations parameter here too. Fixed the missing #ENDINLINE in the template too.

PR #144, opened 2019-08-05
Appengine location

AdrienWalkowiak:
Adding policy to put a constraint on appengine applications location


validator/appengine_location.rego
```diff
@@ -0,0 +1,62 @@
...
+	constraint := input.constraint
+	lib.get_constraint_params(constraint, params)
+
+	asset := input.asset
+
+	# Applies to appengine applications only
+	asset.asset_type == "appengine.googleapis.com/Application"
+
+	# Retrieve the list of allowed locations - default to empty set
+	locations := cast_set(lib.get_default(params, "locations", {}))
```
morgante: If locations are a required param, we shouldn't set a default.

AdrienWalkowiak: updated too. Thanks!

PR #146, opened 2019-08-07
Add tooling for organizing constraint templates.

briantkennedy:
@morgante I've added some initial tooling to handle bundles / tagging the constraint templates.  Since it's still in the works, I haven't added any unit tests (I'll get to that once we confirm this is the right direction).  Please take a look.


morgante:
@briantkennedy What do you think of folding this functionality into config-validator itself? I'm not sure if this needs to be a separate tool.

briantkennedy:
@morgante I think a separate tool is justified since the current config validator is a server daemon while this is more of a CLI.  That said, there's probably a discussion on whether it makes sense for this tool to go here or in the config validator repo.

morgante:
@briantkennedy That's true, but I'm also thinking for certain use cases (ex. CFT Scorecard) it'd be useful to have the bundling functionality integrated so I can merely point at policy-library and say I want to evaluate X bundle. Maybe we could add bundle manager functionality to the config validator package and add a separate CLI entrypoint?

briantkennedy:
@morgante that makes sense, however, putting this into constraint framework means that we will need to make a more general version that's not forseti specific then still create a CLI/library somewhere that handles the forseti specific aspects (mainly related to how we look for the bundles.forsetisecurity.org annotation).

morgante:
@briantkennedy I don't quite follow, I was suggesting we put it in [Forseti Config Validator package](https://github.com/forseti-security/config-validator) which is Forseti-specific, not the constraint framework.

briantkennedy:
Sorry about that, I didn't get enough sleep and mixed up config validator with constraint framework.  I'll create a pull request for config validator shortly.

PR #154, opened 2019-08-12
Constraint documentation (initial POC)

brightjustin:
Create an index of constraints, including the applicable service, a link to the template .yaml file, and a brief description. The table itself is intended to be a mockup of what the end result will look like, as ideally we'll have a script to build this table dynamically in the future. If we want to get something in place before we can design a script, I can create manual entries for the other items in the library.

For now I'd love feedback on the page, and in particular the information in the table. I'm not sure if it's always going to be possible to draw a line between a template and a single service, which is an issue we're dealing with in the Org Policy space as well. I'll also need help in validating the descriptions. 

Is there more information that we'd like to see in the table for an index of this sort?


morgante:
This looks good so far. I think the linkage to individual services makes sense.

My one suggestion would also be to include a column with links to [samples](https://github.com/forseti-security/policy-library/tree/master/samples) of using the template.

brightjustin:
Morgante, that's a good idea. I've added that column. It looks like there's one sample for every template right now, is that right?

Let me know if there's anything else you'd like to see changed or added.

morgante:
Actually some templates have a few samples. Ex. has GCPIAMAllowedBindingsConstraintV1 has 3 samples:
- https://github.com/forseti-security/policy-library/blob/master/samples/iam_blacklist_public.yaml
- https://github.com/forseti-security/policy-library/blob/master/samples/iam_blacklist_role.yaml
- https://github.com/forseti-security/policy-library/blob/master/samples/iam_restrict_role.yaml

AdrienWalkowiak:
any update on this?

AdrienWalkowiak:
@brightjustin @morgante  Any update on this PR?

brightjustin:
I'm planning to wrap up manually creating the index once I'm through with NEXT-related doc releases. In the meantime I'd love any further feedback on the format and contents of this table, so that we can use that as a template for automating its creation.

PR #156, opened 2019-08-16
Updating cloud build integration

AdrienWalkowiak:
Updating builder image to include python3 and pip3 (+yaml module)
Adding target to the Makefile to check if make fmt would change any file in lib/ validator/ 
Adding steps to the cloudbuild file to run make check_samples and make test_format


AdrienWalkowiak:
@morgante can you take a look? We need the builder to be updated with this new Dockerfile for this to work.
We also need to add a trigger in the build project for cloudbuild to trigger on all branches.

PR #158, opened 2019-08-22
Restricted fw rule

AdrienWalkowiak:
This is a test PR at this point. This rule is working and tested in the library, but I need to do more testing against real orgs to figure out if we are having performance issues at scale.

Do not merge it yet


AdrienWalkowiak:
ping @joecheuk ?

morgante:
I'm thinking we should have slightly more complicated semantics to allow for exceptions.

In particular, I'd suggest structuring the props as follows:
```
spec:
  severity: high
  match:
    target: ["organization/*"]
  parameters:
    rules:
    - direction: "INGRESS"
       rule_type: "allowed"
       protocol: "all"
       source_tags:
       - "second-source-tag"
       source_ranges:
        - "192.168.1.1/32"
        - "0.0.0.0/0"
       target_tags:
        - "target-tag"
       source_service_accounts:
        - "cf-gcp-challenge-dev-sa@cf-gcp-challenge-dev.iam.gserviceaccount.com"
    exclude:
    - target_tags: "safe-tag"
```

MartenvanWezel:
Dumb question, but can't we go fully bizlogic on this?

Basically a rego function you can custom define, that takes things as an input and then responds with a boolean true or false?

e.g. 

if martensnetworkparse(params.ip) then fine, else violation
and have martensnetworkparse() contain anything from clever math to just a raw array match.

morgante:
I'd say that's the level where you should just write a custom constraint template though. That's why we made this system flexible enough.

MartenvanWezel:
OK! 

Thing is that in general, many network admin (and in fact the designers of GCP firewalls) still think in the rule,rule,rule,fallback approach. Implementing something in those lines in forseti/CV seems helpful so people don't have to make the translation

(and could potentially open the path to an automated rule importer. I was thinking that in an ideal world, we could use an approved terraform plan as the source of truth for the CV/forseti config, rather than manually creating a config from scratch...

morgante:
That's essentially my recommendation above. Write out the rules you want to alert on, but then write out `exclusion` rules which shouldn't be targeted.

MartenvanWezel:
Just to wrap my head around this: in FW-Land you would write:

deny-based:
- If src=a then allow
- if src=b and port=c then allow
- if srcmask=d.d.d.d/e then allow
[.....]

and the rest deny

In CV/FS land, with your proposal, I imagine you would have to put all the logic in a single statement? (because a stand-alone "deny" rule would have to know about the conditionals that exist but weren't matched)

No?

Not a massive deal but I can imagine that with big corps the amount of FW rules becomes absolutely huge, and the statement to check if it's still correct would be very large, and breakable.. no?


since a violation would be triggered if src=a, even if

morgante:
This policy is not designed to directly mirror firewall rules, but you could certainly write a policy that mirrors that by including the desired conditions in the "exclude" section.

gkowalski-google:
@googlebot I fixed it.

AdrienWalkowiak:
Added support for your proposed change @morgante . I only added the rules: part for now, I will add the exlusion logic in a future PR.

PTAL

policies/templates/gcp_restricted_firewall_rules_v1.yaml
```diff
@@ -0,0 +1,628 @@
...
+        kind: GCPRestrictedFirewallRulesConstraintV1
+        plural: gcprestrictedfirewallrulesconstraintsv1
+      validation:
+        openAPIV3Schema:
+          properties:
+            rules:
+              description: "List of rules to match in order to raise violations. Rule are object describing rules (see below)"
+              type: array
+              items: objects
+              additionalProperties:
```
morgante: I don't think this is quite right for a json schema?

policies/templates/gcp_restricted_firewall_rules_v1.yaml
```diff
@@ -0,0 +1,628 @@
...
+                  default: "any"
+                port:
+                  description: |
+                    "Port of the rules to restrict. This parameter is tighly coupled with the
+                    protocol paramter: the rule will be marked as in violation if the parameter
+                    protocol:port pair matches one of its
+                    protocol:port listed.
+                    Valid values are:
+                        * single number
+                        * ranges (e.g 22-25) - Note: only FW rules with a range greater or equal
```
morgante: I think logic should be that any rule within the provided range would be a violation?

policies/templates/gcp_restricted_firewall_rules_v1.yaml
```diff
@@ -0,0 +1,628 @@
...
+                  description: "Direction of the rules to restrict."
+                  type: string
+                  enum: ["INGRESS", "EGRESS", "any"]
+                  default: "any"
+                rule_type:
+                  description: "Type of the rules to restrict."
+                  type: string
+                  enum: ["allowed", "denied", "any"]
+                  default: "any"
+                port:
```
morgante: Should this be ports?

policies/templates/gcp_restricted_firewall_rules_v1.yaml
AdrienWalkowiak: right, I was not sure how to document these nested fields here. Any suggestion?

policies/templates/gcp_restricted_firewall_rules_v1.yaml
AdrienWalkowiak: The logic of this policy is to let you describe fw rules that should raise a violation. So the way the range parameter works is that if a fw rule range includes the passed range (here) then it raises a violation.

My intention was that, as a user I want to flag all fw rules that match specific criteria, so here I considered that if I want to flag rules with ports 20-30 and and a fw happens to be 10-40, I still flag it since my criteria is true for that rule (it includes 20-30). 

Is it something we should change? The original PRD was only exact matches for ports, even for ranges.

policies/templates/gcp_restricted_firewall_rules_v1.yaml
AdrienWalkowiak: I think so, I matched the name in the original CAI exports but they got updated. Will fix.

policies/templates/gcp_restricted_firewall_rules_v1.yaml
morgante: Good point, that makes sense.

policies/templates/gcp_restricted_firewall_rules_v1.yaml
morgante: https://swagger.io/docs/specification/data-models/data-types/#array

policies/templates/gcp_restricted_firewall_rules_v1.yaml
AdrienWalkowiak: Actually, currently the rule only support one port:protocol pair as a parameter. 

I could add support for more than one but now that we added the "rules: " top-level object it's less needed, I think.

morgante: Got it, let's just make sure the schema is correct then.

AdrienWalkowiak: I updated the schema to fix this (and protocol) and update the properties for rule.

gkowalski-google:
@AdrienWalkowiak Any ETA on when this can be completed/merged? Is there anything I can do to help?

AdrienWalkowiak:
@gkowalski-google  Just waiting on @morgante's review/approval for it.

AdrienWalkowiak:
@morgante , any update on this? Any changes needed?

policies/templates/gcp_restricted_firewall_rules_v1.yaml
```diff
@@ -0,0 +1,627 @@
...
+                      parameter.
+                      Valid values are:
+                          * 'X.X.X.X/XX' - i.e valid CIDR blocks
+                          * 'any' - when set to any, the source ranges of the rule will NOT be a factor
+                                    to determine if the FW rule is restricted
+                          * '*' - all CIDR blocks match: this only checks that a source range exists
+                                  in the restricted FW rule"
+                    type: array
+                    items: string
+                    default: "['any']"
```
morgante: @AdrienWalkowiak Not sure if this should be a string or an array?

```suggestion
                    default: ['any']
```

policies/templates/gcp_restricted_firewall_rules_v1.yaml
```diff
@@ -0,0 +1,627 @@
...
+                properties:
+                  direction:
+                    description: "Direction of the rules to restrict."
+                    type: string
+                    enum: ["INGRESS", "EGRESS", "any"]
+                    default: "any"
+                  rule_type:
+                    description: "Type of the rules to restrict."
+                    type: string
+                    enum: ["allowed", "denied", "any"]
```
morgante: Shouldn't this be `allow` and `deny` to match the standard language?

morgante:
I've approved the PR, just some minor stylistic notes.

policies/templates/gcp_restricted_firewall_rules_v1.yaml
AdrienWalkowiak: indeed - fixed.

policies/templates/gcp_restricted_firewall_rules_v1.yaml
AdrienWalkowiak: Well I matched the CAI terminology here. I could update this to allow or deny but the test for this field would be slightly redundant. Instead of one simple:

```
# test that the key exists in the fw_rule (allowed or denied)
        	fw_rule[lower(rule_type)]
```

we would need to split this into **2 functions** which would look like:

```
rule_type == "allow"
# test that the key exists in the fw_rule
        	fw_rule["allowed"]
```

Is this ok with you?

morgante: If that's the CAI terminology, I'm fine with keeping it as-is.

PR #160, opened 2019-08-26
Brunore/sql maintenance window

BrunoReboul:


BrunoReboul:
@AdrienWalkowiak I commit with the wrong account (no cla) so I create a new PR.
new real test data. the first gen instances do not support a maintenance window and are now filtered out. on second gen instances the maintenanceWindow always exist. day is set to 0 when the maintenance window is not specified

BrunoReboul:
wrong cla

Issue #164, opened 2019-09-03
Update documentation for Policy Library Sync


gkowalski-google:
To support the [Policy Library Sync change](https://github.com/forseti-security/terraform-google-forseti/issues/238), update the documentation for how users can utilize this feature.

PR #165, opened 2019-09-04
Updated user guide for the policy library sync.

gkowalski-google:
Addresses #164. Updated documentation for the policy library sync feature that will be released with Forseti 2.20.0. Related Terraform PR: https://github.com/forseti-security/terraform-google-forseti/pull/239

The updated user guide can be viewed here: https://github.com/gkowalski-google/policy-library/blob/feature/policy-library-sync-documentation/docs/user_guide.md

### DIFF
--- a/docs/user_guide.md
+++ b/docs/user_guide.md
@@ -353,8 +353,6 @@ module to install Forseti. Here is a sample main.tf file modeled mostly from the
 ```
 module "forseti" {
       source  = "terraform-google-modules/forseti/google"
-      version = "~> 1.4"
-
 
       domain             = "yourdomain.com"
       project_id         = "your-forseti-project-id-here"
--- a/docs/user_guide.md
+++ b/docs/user_guide.md
@@ -395,6 +395,9 @@ To get the generated SSH key from Terraform run this command:
 terraform output forseti-server-git-public-key-openssh
 ```
 
+You can view any logs related to this process from Stackdriver Logging by
+searching for `git-sync`.
+
 ### Copy over policy library repository (DEPRECATED)
 
 **NOTE:** This feature has been deprecated and will be removed in a future
--- a/docs/user_guide.md
+++ b/docs/user_guide.md
@@ -14,7 +14,8 @@
   * [For Production Environments](#for-production-environments)
 * [How to Use Forseti Config Validator](#how-to-use-forseti-config-validator)
   * [Install Forseti](#install-forseti)
-  * [Copy over policy library repository](#copy-over-policy-library-repository)
+  * [Policy Library Sync](#policy-library-sync)
+  * [Copy over policy library repository](#copy-over-policy-library-repository-deprecated)
   * [How to change the run frequency of Forseti](#how-to-change-the-run-frequency-of-forseti)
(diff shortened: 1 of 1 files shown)


morgante:
This LGTM, but let's get approval from @joecheuk as well.

We should also wait to merge this until the functionality is actually released.

docs/user_guide.md
```diff
@@ -335,14 +360,48 @@ module "forseti" {
       project_id         = "your-forseti-project-id-here"
       org_id             = "your-org-id-here"
       …
-      config_validator_enabled = true
+      config_validator_enabled      = true
```
joecheuk: I think the TF module version also needs to be updated here `version = "~> 1.4"` *might* not work with the library sync if it's earlier than 4.1.0?

docs/user_guide.md
```diff
@@ -335,14 +360,48 @@ module "forseti" {
...
+`policy_library_sync_enabled` fields. They are not enabled by default; therefore
+you need to explicitly enable them. It is recommended to use the [Policy Library Sync](#policy-library-sync) feature, but not required. If you are using the
+sync, then set `policy_library_repository_url` to the Git URL of your private
+repo.
+
+### Policy Library Sync
+
+Forseti has the ability to clone your private policy library repo and keep it in
+sync (using [git-sync](https://github.com/kubernetes/git-sync)), so that config
+validator will always use the latest policies while scanning. Optionally, you
```
joecheuk: This is not optional to use policy-library sync feature right?

docs/user_guide.md
```diff
@@ -335,14 +360,48 @@ module "forseti" {
...
+for the `policy_library_sync_ssh_known_hosts` field within the main.tf:
+
+```
+ssh-keyscan ${YOUR_GIT_HOST}
+```
+
+After applying the Terraform configuration, you will need to add the generated
+SSH key to the read-only user's account. The SSH key will be provided as an
+output from Terraform. If your private repo is hosted on GitHub, you can
+[follow these steps to add the SSH key to your account](https://help.github.com/en/articles/adding-a-new-ssh-key-to-your-github-account).
```
joecheuk: (non blocking) Instead of adding the ssh key directly to the user's account, is there any way to make it read-only access? Assuming the user is the admin of a few repos, if you were to add the SSH key to the profile directly, I think you can update/delete all the repos under that user?

docs/user_guide.md
gkowalski-google: I don't think it's required. Going forward it will be the only automated way of ensuring Forseti is using the up to date policies, but users could update the policies manually.

docs/user_guide.md
gkowalski-google: The only way to make it read-only is to setup a read-only user within the organization, which is recommended to be done in the [setup section](https://github.com/gkowalski-google/policy-library/blob/feature/policy-library-sync-documentation/docs/user_guide.md#how-to-set-up-constraints-with-policy-library).

docs/user_guide.md
gkowalski-google: Removed this reference, the sample provided in the docs will just use the latest version of the Forseti module.

PR #166, opened 2019-09-05
GCE network interface whitelist policy

charliewolf:


policies/templates/gcp_compute_network_interface_whitelist_v1.yaml
```diff
@@ -0,0 +1,84 @@
...
+  name: gcp-compute-network-interface-whitelist-v1
+spec:
+  crd:
+    spec:
+      names:
+        kind: GCPComputeNetworkInterfaceWhitelistConstraintV1
+        plural: gcpcomputenetworkinterfacewhitelistconstraintsV1
+      validation:
+        openAPIV3Schema:
+            whitelist:
```
AdrienWalkowiak: same here - could you add a comment for this parameter? (is it mandatory/optional, what is its default value + a valid example for it)

policies/templates/gcp_compute_network_interface_whitelist_v1.yaml
```diff
@@ -0,0 +1,86 @@
...
+            
+            	whitelist := lib.get_default(params, "whitelist", [])
+            	allowed_networks := {n | n = whitelist[_]}
+            
+            	access_configs := lib.get_default(interface, "accessConfig", [])
+            
+            	is_external_network := count(access_configs) > 0
+            	is_network_whitelisted := count({full_network_uri} - allowed_networks) == 0
+            
+            	is_external_network == true
```
AdrienWalkowiak: Is there a reason to use a temporary object here? Why not just the test statements directly here (and test for != 0 for L78)?

charliewolf: I think creating an `is_external_network` var helps readability.

AdrienWalkowiak: fair point :)

PR #171, opened 2019-09-13
Bruno/sql backup

BrunoReboul:
This rule checks if sql instances have a backup
It is possible to specify exemptions


BrunoReboul:
@AdrienWalkowiak is this ok for you?

PR #173, opened 2019-09-13
Add more CMEK settings

charliewolf:


charliewolf:
@morgante i think @joecheuk had requested i make it part of the same one, and that's the approach that made most sense to me as well, but i'm open to splitting it out

joecheuk:
There might be some miscommunication but the decision we made was to create a separate one for the new use cases.

AdrienWalkowiak:
Looks like there are conflicts to resolve on this one.

AdrienWalkowiak:
If we are going for a separate constraint then no need to solve the conflict - just close this PR.

ocervell:
What's the rationale for creating a separate rule here ?

charliewolf:
@AdrienWalkowiak  @morgante @joecheuk it is now a separate policy

charliewolf:
@AdrienWalkowiak why did you close this?

AdrienWalkowiak:
I thought you would open a new PR for the new policy and leave this one as-is. Sorry about that.

joecheuk:
@morgante this is now a new template, PTAL

PR #175, opened 2019-09-16
add exemption list to kms

BrunoReboul:
#56 
It is not possible to delete a key ring.
Adding an exemption list enable to mute the violations once fixed
fixture update with 1 more test


BrunoReboul:
@AdrienWalkowiak feedback welcome

PR #177, opened 2019-09-23
Policy for GKE private google access

charliewolf:


policies/templates/gcp_gke_enable_private_google_access_v1.yaml
```diff
@@ -0,0 +1,81 @@
...
+            	"details": metadata,
+            }] {
+            	constraint := input.constraint
+            	asset := input.asset
+            	asset.asset_type == "container.googleapis.com/Cluster"
+            
+            	cluster := asset.resource.data
+            	private_google_access_disabled(cluster)
+            
+            	message := sprintf("Stackdriver monitoring is disabled in cluster %v.", [asset.name])
```
Red-Five: Message doesn't match the rule.

dekuhn:
This rule won't work with Forseti today due to the pagination and the need for this rule to reference data that may not be available. @joecheuk can you work to define how we should annotate this in regards to not usable with Forseti.

I suspect this rule will work fine in Gatekeeper since all the k8s data will all be available at eval time.

morgante:
@dekuhn This certainly wouldn't work in Gatekeeper because it depends on looking at the configuration of the subnets which GKE is running on (information which is never actually exposed to k8s).

dekuhn:
@morgante would this be a useful in CFT Scorecard?

@blueandgold @ryanismert the team should think about how we could run such a scan within Forseti. We will have this challenge with other rules going forward I am sure.

morgante:
@dekuhn It would be, if/when we support referential constraints.

dekuhn:
@blueandgold @charliewolf @joecheuk I don't think we want to accept this rule due to the referential data and the decision by the CV team to no support referential constraints at this time. 

Should we close this PR as it and preserve its details. We could reopen at some point in the future when referential constraints will be supported.

PR #182, opened 2019-10-09
Update example cron job.

blueandgold:
I like to avoid specifying a particular minute, especially at the top of the hour.

### DIFF
--- a/docs/user_guide.md
+++ b/docs/user_guide.md
@@ -427,13 +427,13 @@ desired time format, you can update the run frequency by following the steps
 below:
 
 In main.tf, under module "forseti" include **forseti_run_frequency** and set the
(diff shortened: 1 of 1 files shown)


gkowalski-google:
Do we really need this documented in the policy library? I don't think this or [scaling](https://github.com/forseti-security/policy-library/blob/master/docs/user_guide.md#how-to-handle-scaling-for-large-resource-sets) or [CSCC](https://github.com/forseti-security/policy-library/blob/master/docs/user_guide.md#how-to-handle-scaling-for-large-resource-sets) belong in this documentation/repo. If any of this is not documented on the website, then let's create a ticket to add it and delete it from this guide.

gkowalski-google:
This change is no longer relevant, specific Forseti docs have been moved to the website.

PR #193, opened 2019-10-25
adds a script to create rule templates with CRD based naming

yunus:
Since it is cumbersome to deal with CRD, metadata naming and creating all the rego, template and sample files, I have created a script that creates
- rego file
- template file
- sample file
by using Jinja2.

The templates are filled with vm-external-ip example. When you run `make create_new_rule` it asks for **resource**, **feature**, **version** and **human readable name**.


yunus:
Hey there, Is there anyone taking this? Any feedback?

yunus:
I will check it, looks like a better alternative.

PR #195, opened 2019-10-25
CIS 3.05: Ensure that RSASHA1 is not used for zone-signing key in Cloud DNS DNSSEC

charliewolf:


dekuhn:
@joecheuk PTAL. Thanks

PR #198, opened 2019-10-25
Feature/resource value pattern

AdrienWalkowiak:
New PR to finalized original PR #76 

Adding a fairly generic constraint template that checks any field on an asset against a regex pattern for validation. It started simply as a way to check labels, then evolved as we started writing more policies and needed to check different values and attributes.

Added allowlist/denylist modes.

A sample playground for this constraint can be found [here](https://play.openpolicyagent.org/p/OU9GLH7C4k).


AdrienWalkowiak:
@abombss I am trying to take over your PR (#76) but it looks like the CLA check is failing here. Any idea why?

morgante:
@AdrienWalkowiak The error is this PR contains commits authored by @abombss. We can manually override though.

AdrienWalkowiak:
Got the green light from @abombss on a separate email thread

abombss:
You have my approval, let me know if I need to actually do anything.

abombss:
@googlebot I consent.

hshin-g:
@AdrienWalkowiak what is the current status of this PR? Is this something that can be merged in? Looks like this can address the allowlist part of the Forseti Kubernetes Engine scanner (issue #269 ) and once this is merged in I can start to work on the denylist part if that works.

AdrienWalkowiak:
@morgante or @t12g could you review this one?

hshin-g:
@morgante @briantkennedy @t12g Could we get this PR reviewed? Thanks!

validator/test/fixtures/resource_value_pattern/constraints/data.yaml
```diff
@@ -0,0 +1,123 @@
...
+        gcp:
+          target: ["organization/*"]
+          exclude: [] # optional, default is no exclusions
+      parameters:
+        mode: allowlist
+        asset_types:
+          - "cloudresourcemanager.googleapis.com/Project"
+        field_name: "labels.billing_id"
+        pattern: "^\\d{4}$"
+        required: true
```
briantkennedy, line 32: "optional: true" may have better UX since required is then implied in absence of the optional param.

validator/test/fixtures/resource_value_pattern/constraints/data.yaml
```diff
@@ -0,0 +1,123 @@
...
+          target: ["organization/*"]
+          exclude: [] # optional, default is no exclusions
+      parameters:
+        mode: allowlist
+        asset_types:
+          - "cloudresourcemanager.googleapis.com/Project"
+          - "compute.googleapis.com/Instance"
+        field_name: "labels.billing_id"
+        required: false
+        pattern: "^\\d{4}$"
```
briantkennedy, line 51: Is it possible that users might want to filter for and against something?  Rather than "mode" and "pattern", would "allow" and "deny" be more flexible?  Is it possible they might want to have a list of allowable patterns?

hshin-g: That's an interesting idea. @AdrienWalkowiak as the OG of this PR, what are your thoughts?

validator/test/fixtures/resource_value_pattern/constraints/data.yaml
hshin-g: That makes sense- have changed to `optional`. Thanks!

PR #202, opened 2019-10-28
CIS 3.06: Ensure that SSH access is restricted from the internet

charliewolf:


AdrienWalkowiak:
This looks like an edge case of the restricted firewall rule policy I wrote (#158 ).

@morgante how do we want to handle this? Is it ok to have "overlapping" policies if it makes it easier for users to implement CIS requirements?

morgante:
I would lean towards *not* having duplicative policies if possible.

Instead, we can/should attach the CIS annotations to *samples* of the modules. So we could have a sample of the firewall constraint w/ the CIS annotation.

PR #212, opened 2019-11-02
adds regex functionality to vm external ip detection

yunus:
Adds regex based matching to external IP constraints. Now it is possible to whitelist all VMs with a naming convention such as ".*frontend.*"

### DIFF
--- a/validator/vm_external_ip.rego
+++ b/validator/vm_external_ip.rego
@@ -19,7 +19,7 @@ package templates.gcp.GCPComputeExternalIpAccessConstraintV1
 import data.validator.gcp.lib as lib
 
 ###########################
-# Find Whitelist Violations
+# Find Whitelist/Blacklist Violations
 ###########################
 deny[{
 	"msg": message,
@@ -36,10 +36,11 @@ deny[{
 	count(access_config) > 0
 
 	# Check if instance is in blacklist/whitelist
+	match_mode := lib.get_default(params, "match_mode", "exact")
 	target_instances := params.instances
-	matches := {asset.name} & cast_set(target_instances)
-	target_instance_match_count(params.mode, desired_count)
-	count(matches) == desired_count
+	trace(sprintf("asset name:%v, target_instances: %v, mode: %v, match_mode: %v", [asset.name, target_instances, params.mode, match_mode]))
+
+	instance_name_targeted(asset.name, target_instances, params.mode, match_mode)
 
 	message := sprintf("%v is not allowed to have an external IP.", [asset.name])
 	metadata := {"access_config": access_config}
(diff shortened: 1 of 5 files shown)


yunus:
I don't see what has failed and of course rune in my own laptop :). Could you please give me viewer access to the cloud build pipeline?

AdrienWalkowiak:
Loos like a formatting issue. Make sure to use make build or make fmt before creating a PR.

I gave you access to the Build project for future needs :)

yunus:
Thanks @AdrienWalkowiak . Now it passes

validator/vm_external_ip_test.rego
```diff
@@ -32,16 +32,16 @@ find_violations[violation] {
 	violation := issues[_]
 }
 
-# Test logic for whitelisting/blacklisting
-test_target_instance_match_count_whitelist {
-	target_instance_match_count("whitelist", match_count)
-	match_count = 0
-}
+# # Test logic for whitelisting/blacklisting
```
morgante: Let's not comment out tests — either remove them entirely if they're no longer applicable or update to use new semantics.

validator/test/fixtures/vm_external_ip/constraints/forbid_external_ip_blacklist/data.yaml
```diff
@@ -21,4 +21,4 @@ spec:
   parameters:
     mode: "blacklist"
     instances:
-      - //compute.googleapis.com/projects/test-project/zones/us-east1-b/instances/vm-external-ip
+      - ^//compute.googleapis.com/projects/test-project/zones/us-east1-b/instances/vm-external-ip$
```
morgante: Do all instance names have to be regex now? I'd prefer if they didn't and we either (a) automatically found ones which are regex based or (b) specially indicated that the name is a regex. Otherwise everyone will have to worry about making their hard-coded names match a proper regex.

validator/vm_external_ip.rego
```diff
@@ -36,10 +36,8 @@ deny[{
 	count(access_config) > 0
 
 	# Check if instance is in blacklist/whitelist
-	target_instances := params.instances
-	matches := {asset.name} & cast_set(target_instances)
-	target_instance_match_count(params.mode, desired_count)
-	count(matches) == desired_count
+trace(	sprintf("asset: %v, instances: %v, mode: %v", [asset.name, params.instances, params.mode]))
+	any_name_match(asset.name, params.instances, params.mode)
```
morgante: Could we call this something like instance_name_targeted?

validator/vm_external_ip_test.rego
yunus: I have created unit tests for matching functions.

validator/test/fixtures/vm_external_ip/constraints/forbid_external_ip_blacklist/data.yaml
yunus: I have created 2 different "match_modes", exact and regex. "Exact" is default, so existing templates do not need any new changes. 

Automatic detection of regex usage is problematic.  I can search for *, ^, $ like symbols but still there are many other valid regex symbols or patterns.

validator/vm_external_ip.rego
yunus: changed the function name

samples/vm_external_ip.yaml
```diff
@@ -22,3 +22,5 @@ spec:
     mode: "whitelist"
     instances:
       - //compute.googleapis.com/projects/test-project/zones/us-east1-b/instances/vm-external-ip-two
+      # you may also provide regex expressions, like:
```
morgante: Since match_mode has to be changed to allow regular expressions, can we provide a separate sample with regex support?

yunus: Sorry I missed this. I added example. I have also updated the template file. match_mode was missing there

PR #213, opened 2019-11-18
Add policy for resource naming convention

yunus:
Enables security teams to create naming policies for the resources. For each resource, you are able to define different and multiple naming patterns

### DIFF
--- a/validator/gcp_enforce_naming.rego
+++ b/validator/gcp_enforce_naming.rego
@@ -1,0 +1,67 @@
+#
+# Copyright 2019 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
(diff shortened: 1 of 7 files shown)


validator/gcp_enforce_naming.rego
```diff
@@ -0,0 +1,67 @@
...
+	# we are interested in the last part of the name, 
+	# the rest is resource address
+	split_name := split(asset_full_name, "/")
+	last_index := count(split_name) - 1
+	name := split_name[last_index]
+}
+
+check_asset_and_return_its_rules(asset_type, naming_rules) = patterns {
+	some i
+	naming_rules[i].resource == asset_type
```
morgante: Not sure if we need `i` here. Could we try this?

```suggestion
  rule = naming_rules[_]
  rule.resource == asset_type
  patterns = rule.patterns
```

PR #222, opened 2019-12-04
Adding support for bigQuery datasets, tables and views

AdrienWalkowiak:
Adding support for bigQuery datasets, tables and views to the default resource type list + adding tests on these resources.

Note: Views's type in CAI export is actually the same as Tables (with extra fields).


validator/enforce_labels_test.rego
```diff
@@ -378,3 +387,31 @@ test_enforce_label_dataproc_violations {
 	is_string(violation.msg)
 	is_object(violation.details)
 }
+
+#### Testing for BQ resources (Datasets, Tables and Views)
+# Confirm exactly 12 violations were found
+# 2 datasets have violations -both of them have 2 violations (missing or wrong labels)
```
xingao267: s/-both/- both

AdrienWalkowiak: fixed

AdrienWalkowiak:
anyone to review this PR?

PR #226, opened 2019-12-19
Added template for restricting default networks

onetwopunch:
It is a best practice not to create the default network when creating a project. In fact in project-factory, this setting, `auto_create_network` is disabled by default. This template, when instantiated, will alert for all projects that have default networks.


onetwopunch:
@AdrienWalkowiak would you be able to review this when you have a moment?

policies/templates/gcp_network_restrict_default_v1.yaml
```diff
@@ -0,0 +1,62 @@
...
+metadata:
+  name: gcp-network-restrict-default-v1
+spec:
+  crd:
+    spec:
+      names:
+        kind: GCPNetworkRestrictDefaultV1
+      validation:
+        openAPIV3Schema:
+          mode:
```
AdrienWalkowiak, line 26: looks like this is never used - could you remove it?

policies/templates/gcp_network_restrict_default_v1.yaml
```diff
@@ -0,0 +1,62 @@
...
+            package templates.gcp.GCPNetworkRestrictDefaultV1
+            
+            import data.validator.gcp.lib as lib
+            
+            deny[{
+            	"msg": message,
+            	"details": metadata,
+            }] {
+            	a := input.asset
+            	count({a.asset_type} & {"compute.googleapis.com/Network", "google.compute.Network"}) == 1
```
AdrienWalkowiak, line 56: Why this test? Is it for backward compatibility?

onetwopunch: Probably copypasta error.

PR #231, opened 2019-12-19
[BLOCKED] CIS 7: Enable workload identity

charliewolf:
CAI does not currently export this field, but I am using the field name listed in the [API docs](https://cloud.google.com/kubernetes-engine/docs/reference/rest/v1beta1/projects.locations.clusters#resource-cluster) (`workloadIdentityConfig.identityNamespace`) and this should work as soon as CAI adds that field.


validator/gke_enable_workload_identity.rego
```diff
@@ -0,0 +1,35 @@
...
+deny[{
+	"msg": message,
+	"details": metadata,
+}] {
+	constraint := input.constraint
+	asset := input.asset
+	asset.asset_type == "container.googleapis.com/Cluster"
+
+	cluster := asset.resource.data
+	identity_namespace := lib.get_default(lib.get_default(cluster, "workloadIdentityConfig", {}), "identityNamespace", null)
```
mmontan: CAI added the field, but called it "workloadPool" rather than identityNamespace.

"workloadIdentityConfig": {
  "workloadPool": "mmontan-cloud.svc.id.goog"
 },

On the other side, Terraform Validator is going to output the field as "identityNamespace"

It would be better to support both, so we can use the constraint across environments.

validator/test/fixtures/gke_enable_workload_identity/assets/data.json
```diff
@@ -0,0 +1,162 @@
...
+              }
+            ],
+            "selfLink": "https://container.googleapis.com/v1/projects/pso-cicd8/zones/us-west1-b/clusters/canary-west",
+            "servicesIpv4Cidr": "10.47.240.0/20",
+            "status": "RUNNING",
+            "subnetwork": "default",
+            "workloadIdentityConfig": {
+                "identityNamespace": "test-project-123-namespace"
+            },
+            "zone": "us-west1-b"
```
mmontan: We should also have a cluster to test with the following:

      "workloadIdentityConfig": {
        "workloadPool": "test-project-123.svc.id.goog"
      },

PR #240, opened 2019-12-30
CIS 1.07 & 1.09: Separation of duties

charliewolf:


samples/kms_role_separation.yaml
```diff
@@ -0,0 +1,18 @@
...
+  name: ensure_service_account_role_separation
+  annotations:
+    description: Prevent public users from having access to both administer and user service accounts at the same time
+    # This constraint is not certified by CIS.
+    bundles.validator.forsetisecurity.org/cis-v1.1: 1.09
+spec:
+  severity: high
+  parameters:
+    roles:
+      - - roles/cloudkms.admin
```
joecheuk: why is the extra dash needed here?

e.g.

```
    roles:
      - - roles/cloudkms.admin
        - roles/cloudkms.cryptoKeyDecrypter
```

instead of 

```
    roles:
      - roles/cloudkms.admin
        roles/cloudkms.cryptoKeyDecrypter
```

charliewolf: @joecheuk because it's a list of lists of roles, not a list of roles. each list is a set that the exclusion is calculated for.

dekuhn:
@gkowalski-google can you please review and help Charlie get this merged. I think he has addressed Joe's questions/feedback.

PR #244, opened 2019-12-30
CIS 4.01: Ensure that instances are not configured to use the default…

charliewolf:
… service account with full access to all Cloud APIs


validator/compute_default_service_account.rego
```diff
@@ -0,0 +1,36 @@
...
+	constraint := input.constraint
+	asset := input.asset
+
+	asset.asset_type == "compute.googleapis.com/Instance"
+	service_account := asset.resource.data.serviceAccount[_]
+	scope := service_account.scope[_]
+
+	scope == "https://www.googleapis.com/auth/cloud-platform"
+
+	message := sprintf("Instance %v has default service account with full access to all Cloud APIs.", [asset.name])
```
katze120: Looks like this block only checks scope but not service account

Suggest to add `re_match("-compute@developer.gserviceaccount.com$", service_account.email)` to check for default compute engine service account 

If we use custom service account, scope will always be full scope, which is actually recommended: https://cloud.google.com/compute/docs/access/create-enable-service-accounts-for-instances#best_practices

joecheuk: Thanks for the recommendation! @charliewolf what do you think about this strategy?

charliewolf: @joecheuk So in my test CAI data i was not seeing service accounts with names like that. @katze120 Do you have an example asset JSON you could share?

dekuhn:
@katze120 or @gkowalski-google can you help with data needed by Charlie base on the code review feedback.

PR #246, opened 2020-01-02
constraint and template to check firewall logs

katze120:


validator/network_enable_firewall_logs.rego
```diff
@@ -0,0 +1,33 @@
...
+
+deny[{
+	"msg": message,
+	"details": metadata,
+}] {
+	constraint := input.constraint
+	asset := input.asset
+	asset.asset_type == "compute.googleapis.com/Firewall"
+
+	asset.resource.data.logConfig.enable == false
```
morgante: Is logConfig always present?

katze120: In the 3 orgs I tested, logConfig is there for all firewall rules, but I updated the rego to set a default nevertheless.

PR #247, opened 2020-01-02
Update sample firewall constraints to match CIS benchmark, and other minor fixes

katze120:


samples/restrict_fw_rules_rdp_open.yaml
```diff
@@ -17,7 +17,7 @@
 apiVersion: constraints.gatekeeper.sh/v1alpha1
 kind: GCPRestrictedFirewallRulesConstraintV1
 metadata:
-  name: restrict-firewall-rule-ssh-open
+  name: restrict_firewall_rule_rdp_open
```
morgante: Why switch to underscore? Kubernetes names can only use hyphens.

katze120: Because majority of constraint names  (66 out of 75) in other samples are using underscore, and I want to make things consistent so that the report will look nicer. Are we going to change all to hyphens in order to be consistent with gatekeeper?

morgante: Yes, I think we should be moving in the other direction (updating other constraint samples to use hyphens).

katze120: ok!

katze120:
A side note: is it possible to update the opa version in gcr.io/config-validator/make? say, v0.16.0?

Currently 'make build' will reformat some times with extra blanks, example: https://github.com/forseti-security/policy-library/blob/master/validator/restricted_firewall_rules.rego#L72 
Below are all the files with similar blanks.

```
	   policies/templates/gcp_allowed_resource_types_v1.yaml
	   policies/templates/gcp_always_violates_v1.yaml
	   policies/templates/gcp_appengine_location_v1.yaml
	   policies/templates/gcp_dataproc_location_v1.yaml
	   policies/templates/gcp_enforce_labels_v1.yaml
	   policies/templates/gcp_restricted_firewall_rules_v1.yaml
	   policies/templates/gcp_spanner_location_v1.yaml
	   validator/allowed_resource_types.rego
	   validator/appengine_location.rego
	   validator/dataproc_location.rego
	   validator/enforce_labels.rego
	   validator/restricted_firewall_rules.rego
	   validator/spanner_location.rego
```

morgante:
> A side note: is it possible to update the opa version in gcr.io/config-validator/make? say, v0.16.0?

Absolutely, please feel free to open a PR: https://github.com/forseti-security/policy-library/blob/master/cloudbuild/Dockerfile

PR #250, opened 2020-01-09
rego format fixes with latest opa version

katze120:
Rego format fixes with opa version upgraded from v0.13.2 to v0.16.0
This PR will now fail the Cloud Build CI test but should pass after #249 is merged


morgante:
@katze120 Even after merging #249 it looks like this is still failing.

PR #252, opened 2020-01-15
Fix some constraint templates

briantkennedy:
The constraint framework migration made FCV pickier about the input since it's actually looking at the fields.  There were some issues in the existing policy library constraint templates that this fixes.

### DIFF
--- a/policies/templates/gcp_app_service_versions.yaml
+++ b/policies/templates/gcp_app_service_versions.yaml
@@ -28,7 +28,6 @@ spec:
               type: integer
               description: "The maxinum number of simultaneous versions
                Defaults to 2"
-              default: 2
   targets:
    validation.gcp.forsetisecurity.org:
       rego: | #INLINE("validator/app_service_versions.rego")
--- a/policies/templates/gcp_vpc_sc_ip_range_v1.yaml
+++ b/policies/templates/gcp_vpc_sc_ip_range_v1.yaml
@@ -28,8 +28,8 @@ spec:
             maximum_cidr_size:
               type: integer
               description: "Maximum size in CIDR notation for an IP network"
-              minimum: 0
-              maximum: 32
+              #minimum: 0
+              #maximum: 32
   targets:
     validation.gcp.forsetisecurity.org:
       rego: | #INLINE("validator/vpc_sc_ip_range.rego")
--- a/validator/gke_enable_stackdriver_kubernetes_engine_monitoring_test.rego
+++ b/validator/gke_enable_stackdriver_kubernetes_engine_monitoring_test.rego
@@ -14,7 +14,7 @@
 # limitations under the License.
 #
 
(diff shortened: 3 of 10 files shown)


policies/templates/gcp_app_service_versions.yaml
```diff
@@ -28,7 +28,6 @@ spec:
               type: integer
               description: "The maxinum number of simultaneous versions
                Defaults to 2"
-              default: 2
```
morgante: Why can't we have a default here?

policies/templates/gcp_vpc_sc_ip_range_v1.yaml
```diff
@@ -28,8 +28,8 @@ spec:
             maximum_cidr_size:
               type: integer
               description: "Maximum size in CIDR notation for an IP network"
-              minimum: 0
```
morgante: Why aren't these supported?

validator/gke_enable_stackdriver_kubernetes_engine_monitoring_test.rego
```diff
@@ -14,7 +14,7 @@
 # limitations under the License.
 #
 
-package templates.gcp.GCPGKEEnableStackdriverKubernetesEngineMonitoringConstraintV1
```
morgante: Any particular reason to remove the `constraint` from the name?

policies/templates/gcp_vpc_sc_ip_range_v1.yaml
briantkennedy: I believe it's a bug in the K8S yaml parser.

validator/gke_enable_stackdriver_kubernetes_engine_monitoring_test.rego
briantkennedy: See commit descriptions

policies/templates/gcp_app_service_versions.yaml
briantkennedy: Constraint framework doesn't allow it.

briantkennedy:
Please see commit descriptions for more info.

policies/templates/gcp_app_service_versions.yaml
morgante: Got it. Can we open an issue/fix upstream, since it seems like something CF should allow?

briantkennedy: Filed https://github.com/open-policy-agent/frameworks/issues/70

briantkennedy: Also, keep in mind that FCV never actually did anything with that field, the change here is that rather than ignoring the default, FCV now errors out.

morgante: Right, I think we can remove for now until that's fixed.

PR #254, opened 2020-01-22
Fix openapi schema and constraints.

briantkennedy:
Some of the sample constraints were not conforming to the openapi
schema.  Also, two templates had errors in the openapi spec portion.

### DIFF
--- a/policies/templates/gcp_restricted_firewall_rules_v1.yaml
+++ b/policies/templates/gcp_restricted_firewall_rules_v1.yaml
@@ -36,7 +36,7 @@ spec:
               description: "List of rules to match in order to raise violations. Rule are object describing rules (see below)"
               type: array
               items:
-                type: objects
+                type: object
                 properties:
                   direction:
                     description: "Direction of the rules to restrict."
@@ -149,8 +149,8 @@ spec:
                     items: string
                   enabled:
                     description: "Status of the rules to restrict (enabled vs disabled)."
-                    type: boolean or string
-                    enum: [true, false, "any"]
+                    type: string
+                    enum: ["true", "false", "any"]
           required: [ "rules" ]
   targets:
     validation.gcp.forsetisecurity.org:
--- a/policies/templates/gcp_vpc_sc_project_perimeter_v2.yaml
+++ b/policies/templates/gcp_vpc_sc_project_perimeter_v2.yaml
@@ -1,0 +1,100 @@
+# Copyright 2019 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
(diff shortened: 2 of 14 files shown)


policies/templates/gcp_vpc_sc_project_perimeter_v1.yaml
```diff
@@ -39,8 +39,8 @@ spec:
               description: "Array of service perimeters that will be allowed/denied (based on mode)
               for this project."
             project_id:
-              type: string
-              description: "The project ID you are matching"
+              type: number
```
morgante: Are you sure this is meant to be project number and not project ID?

briantkennedy: yes, given the samples/*.yaml and line 83 of this template, it's definitely number.

morgante: Got it. In that case, could we also change the field name?

briantkennedy: Unfortunately that's a little trickier if we already have customers using this since we would have to provide legacy support (likely in a rego workaround).

morgante: In that case, could we add a v2 template (with these fixes) while archiving the v1 version with original name?

I'd like to avoid calling a project number a project ID in the long term. It's a very confusing UX.

briantkennedy: done

briantkennedy:
I put the old ones in samples/legacy, is that what you're asking for?

morgante:
> I put the old ones in samples/legacy, is that what you're asking for?

We should also have copies for the new (v2) version under `samples/` with `project_id` changed to `project_number` - otherwise the linter complains.

briantkennedy:
Are these not suitable?  Am I missing something? https://github.com/forseti-security/policy-library/pull/254/files#diff-d091b094bfc55053567008c07eafb352

briantkennedy:
Here's the old ones: https://github.com/forseti-security/policy-library/pull/254/files#diff-b066b04f84b5c40ccfccac670b82a996

briantkennedy:
Also, keep in mind Git determines which files are moved / changed, not me, so I can't go and alter what it thinks I did.

morgante:
Ah, I think I see the issue. It looks like the linter isn't recursively parsing the `samples` directory to find the old samples under `legacy/`:

```
No sample found for template GCPVPCSCProjectPerimeterConstraintV1
```

briantkennedy:
This should be fixed now.

PR #276, opened 2020-02-26
GKE Cluster Location Constraint v2

### LINKED ISSUE #270: [FORSETI] Location scanner functionality
Related to Forseti Security Issue [#3624](https://github.com/forseti-security/forseti-security/issues/3624)

### Scanner

Location Scanner

**Description:** Allow customers to ensure their resources are located only in the intended locations. Set guards around locations as part of automated project deployment.

### Python Violations Summary
There are two types (modes) that can generate violations:

1) **Whitelist** - generate a violation if location value extracted is NOT in the list defined in location_rules.yaml file.

2) **Blacklist** - generate a violation if location value extracted IS in the listed defined in location_rules.yaml file.

**Whitelist**

For each rule that a user has defined in the location_rules.yaml file, there is a “locations” field that specifies what location strings are allowed for the resources specified in “applies_to” field. For each resource object that matches the “applies_to” field, if the “locations” field of the resource object does NOT contain the string defined in location_rules.yaml file, generate a violation.

**Blacklist**

For each rule that a user has defined in the location_rules.yaml file, there is a “locations” field that specifies what location strings are allowed for the resources specified in “applies_to” field. For each resource object that matches the “applies_to” field, if the “locations” field of the resource object DOES contain the string defined in location_rules.yaml file, generate a violation.

### 1:1 Rego | Python Parity Status
The Python scanner has the ability to whitelist/blacklist, which the Rego rules seem to have. However, the Python scanner is also able to only scan for resources with specific ids, which the Rego rules do NOT seem to have. Some rules DO have exemptions (bucket, cloudsqlinstance, dataset), which serve to “ignore” resources with specific ids (although the two functions are not entirely the same, the exemptions do provide a workaround). There is no Rego rule for Compute Instance location.

### Rego work to do

Create or edit existing Rego rules to do the following:
- gcp_storage_location_v1.yaml
- gcp_sql_location_v1.yaml
- gcp_gke_cluster_location_v1.yaml
- gcp_bq_dataset_location_v1.yaml
- gcp_appengine_location_v1.yaml
- gcp_dataproc_location_v1.yaml
- gcp_spanner_location_v1.yaml
- gcp_compute_zone_v1.yaml (Instance types only have "zone" parameter)

1) Allow ""exemption"" option for the following resources -->  currently this feature does not exist.
    - Cluster

2) Allow ""denylist"" option for Cluster resource --> currently this feature does not exist.

hshin-g:
Resolves #270 

This PR adds in the following features to the GKE Cluster Location rego constraint (v1 [here](https://github.com/forseti-security/policy-library/blob/master/policies/templates/gcp_gke_cluster_location_v1.yaml)):
- allowlist/denylist mode
- exemptions
- more tests


hshin-g:
@morgante This changes the "allowed_locations" parameter to "locations" to better fit in with the allowlist/denylist syntax. Would that be considered incompatible?

morgante:
@hshin-g Ah I didn't catch that. Yes you're right, that's an incompatible change.

samples/gke_cluster_location_v2.yaml
```diff
@@ -0,0 +1,28 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+apiVersion: constraints.gatekeeper.sh/v1alpha1
+kind: GKEClusterLocationConstraintV2
+metadata:
+  name: gke_cluster_v2_denied_locations
```
morgante: We want to switch to dashes in names, to match k8s style.

validator/gke_cluster_location_v2_test.rego
```diff
@@ -0,0 +1,249 @@
+#
```
morgante: We don't need to calls this file `_v2` because only the latest version of constraints is maintained/actively updated.

samples/gke_cluster_location_v2.yaml
hshin-g: Ack. PTAL.

Issue #271, opened 2020-02-18
[FORSETI] Retention Policy scanner functionality


hshin-g:
Related to Forseti Security Issue [#3624](https://github.com/forseti-security/forseti-security/issues/3624)

### Scanner

Retention Policy Scanner

**Description:** Allow customers to ensure the retention policies on their resources are set as intended.

### Python Violations Summary
There are two resources that each have two checks that can generate violations:

**Bucket**
1) Maximum_retention - generate a violation:

    - If the “age” field is greater than the maximum_retention value found in the retention_policy_rules.yaml file.

2) Minimum_retention- generate a violation:

   - If the “age” field is less than the minimum_retention value found in the retention_policy_rules.yaml file.

   - If the “createdBefore” field exists under the “condition” field, calculate the difference in days between the day the scanner is run and the “createdBefore” field. If the difference in days is LESS than the minimum_retention field.

   - If the “numNewerVersions” field exists under the “condition” field AND it has a value less than 1.

**Bigquery table**

1) Maximum_retention- generate a violation:

   - If the “expirationTime” field does NOT exist.

   - If the “expirationTime” field exists and “creationTime” field exists, get the difference between the expirationTime and the creationTime. If this difference is GREATER than the maximum_retention value in retention_policy_rules.yaml file.

2) Minimum_retention - generate a violation:

   - If the “expirationTime” field exists and “creationTime” field exists, get the difference between the expirationTime and the creationTime. If this difference is LESS than the minimum_retention value in retention_policy_rules.yaml file.

### 1:1 Rego | Python Parity Status
Cannot seem to find any Rego rules that pertain to this specific functionality. Will need to write the Rego rules for this.

### Rego work to do

Create Rego rule(s) to do the following:

1) Check for ""maximum_retention"" and ""minimum_retention"" for Bucket resource --> currently this rule does not exist.

2) Check for ""maximum_retention"" and ""minimum_retention"" for Bigquery Table resource --> currently this rule does not exist.

hshin-g:
@morgante : Will have to write these as they do not currently existing.

PR #281, opened 2020-03-02
BigQuery Table Retention Constraint v1

hshin-g:
BigQuery Table constraint for #271

This PR creates a BigQuery Table Retention constraint that checks for the retention policy of a BigQuery table and compares the user defined number of days the retention of a BigQuery table should be against what it actually is, utilizing the creationTime and expirationTime fields found in the BigQuery table resource.

This constraint provides the following features:

1. **minimum_retention_days**
- minimum_retention_days identifies the minimum number of days the retention of a BigQuery table should be. 
  - Generate a violation if the resource retention is less than the minimum number of retention days allowed.
2. **maximum_retention_days**
- maximum_retention_days identifies the maximum number of days the retention of a BigQuery table can be.
  - Generate a violation if the resource retention is greater than the maximum number of retention days allowed.
  - Generate a violation if expirationTime does not exist when looking at the maximum retention.
3. **Exemptions**
- BigQuery tables to exempt from retention restriction.

NOTE: Can use values for both minimum_retention_days and maximum_retention_days to define a range of days that the resource retention should be within.

Also included in this PR:

- Sample constraint
- Tests

### DIFF
--- a/samples/bigquery_table_retention.yaml
+++ b/samples/bigquery_table_retention.yaml
@@ -21,8 +21,6 @@ spec:
   match:
     target: ["organization/*"]
   parameters:
-    retention_type: "minimum_maximum"
-    retention_days:
-      - 100
-      - 200
+    minimum_retention_days: 100
+    maximum_retention_days: 200
     exemptions: []
--- a/samples/bigquery_table_retention.yaml
+++ b/samples/bigquery_table_retention.yaml
@@ -1,0 +1,28 @@
+# Copyright 2020 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+apiVersion: constraints.gatekeeper.sh/v1alpha1
+kind: GCPBigQueryTableRetentionConstraintV1
+metadata:
+  name: bq_table_minimum_maximum_retention
+spec:
+  severity: high
+  match:
+    target: ["organization/*"]
+  parameters:
+    retention_type: "minimum_maximum"
+    retention_days:
+      - 100
+      - 200
+    exemptions: []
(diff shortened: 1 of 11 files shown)


samples/bigquery_table_retention.yaml
```diff
@@ -0,0 +1,28 @@
...
+kind: GCPBigQueryTableRetentionConstraintV1
+metadata:
+  name: bq_table_minimum_maximum_retention
+spec:
+  severity: high
+  match:
+    target: ["organization/*"]
+  parameters:
+    retention_type: "minimum_maximum"
+    retention_days:
```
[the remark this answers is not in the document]
hshin-g: That's a really great point @briantkennedy. Have updated the constraints accordingly- PTAL!

validator/bigquery_table_retention.rego
```diff
@@ -0,0 +1,84 @@
...
+
+	diff := expiration_time - creation_time
+	diff > retention_days_ms
+}
+
+# If expirationTime does not exist when looking at the maximum retention, generate a violation.
+get_diff(asset, minimum_retention_days, maximum_retention_days) {
+	maximum_retention_days != ""
+	get_expiration_time := object.get(asset.resource.data, "expirationTime", "")
+	get_expiration_time == ""
```
briantkennedy: interesting scenario, if a user specifies a maxium, but no minimum, should this cause  a violation?  From the perspective of least surprise, I think that no retention time should be OK here given that the retention time is effectively 0 which is less than the maximum constraint.

validator/bigquery_table_retention.rego
```diff
@@ -0,0 +1,84 @@
...
+	asset.asset_type == "bigquery.googleapis.com/Table"
+
+	# Check if resource is in exempt list
+	exempt_list := params.exemptions
+	matches := {asset.name} & cast_set(exempt_list)
+	count(matches) == 0
+
+	get_diff(asset, min_retention_days, max_retention_days)
+
+	message := sprintf("BigQuery table %v has a retention policy violation.", [asset.name])
```
briantkennedy: is there a way to surface the violation in the message, eg, x < min or max < x (not necessary if this is overly complicated).

hshin-g: I've added a string as an output that will hopefully shed some better light- PTAL!

validator/bigquery_table_retention.rego
hshin-g: I think if the user specifies that there should be a maximum retention, then if there is no expiration date, it should generate a violation since that is a resource whose retention is not being tracked / can potentially overshoot the user defined maximum retention.

briantkennedy: oh, good point, I misunderstood that it defaults to infinity, this seems like the right thing to do then.

PR #282, opened 2020-03-05
Added optional support for cos_containerd.

mmontan:
Modified the gke_container_optimized_os to support optionally specifying
COS_CONTAINERD images as allowed. COS_CONTAINERD is necessary for GKE
and gVisor.

### DIFF
--- a/policies/templates/gcp_gke_container_optimized_os.yaml
+++ b/policies/templates/gcp_gke_container_optimized_os.yaml
@@ -57,11 +57,13 @@ spec:
             	constraint := input.constraint
             	asset := input.asset
             	asset.asset_type == "container.googleapis.com/Cluster"
+            	lib.get_constraint_params(constraint, params)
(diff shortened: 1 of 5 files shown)


policies/templates/gcp_gke_container_optimized_os.yaml
```diff
@@ -57,11 +57,13 @@ spec:
             	constraint := input.constraint
             	asset := input.asset
             	asset.asset_type == "container.googleapis.com/Cluster"
+            	lib.get_constraint_params(constraint, params)
+            	containerd_allowed = lib.get_default(params, "cos_containerd_allowed", false)
```
morgante: The new param should be added to the openAPI spec for the template (line 29).

mmontan: Added. Thanks!

PR #283, opened 2020-03-05
Constraint for GKE Shielded Nodes

mmontan:
The constraint enforces that GKE clusters are created with shielded nodes, and that both secure boot and integrity logging are enabled on all node pools.

### DIFF
--- a/validator/gke_enable_shielded_nodes.rego
+++ b/validator/gke_enable_shielded_nodes.rego
@@ -32,7 +32,7 @@ deny[{
 
 	# We fail if either the cluster doesn't have shielded vms, or
 	# one of the node pools doesn't have the proper settings.
-	cluster_shielded_nodes(cluster, node_pool)
+	cluster_shielded_nodes_disabled(cluster, node_pool)
 
 	message := sprintf("Cluster %v doesn't use shielded VMs with integrity and secure boot in node pool %v.", [asset.name, node_pool.name])
 	metadata := {"resource": asset.name}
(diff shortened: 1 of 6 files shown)


validator/gke_enable_shielded_nodes.rego
```diff
@@ -0,0 +1,67 @@
...
+	message := sprintf("Cluster %v doesn't use shielded VMs with integrity and secure boot in node pool %v.", [asset.name, node_pool.name])
+	metadata := {"resource": asset.name}
+}
+
+###########################
+# Rule Utilities
+###########################
+
+# Checks the Shielded Nodes setting in the cluster configuration.
+cluster_shielded_nodes(cluster, node_pool) {
```
morgante: I think the naming on these functions should be clearer to show they are looking *for* violations (ie. return true when the cluster or node pool is *not* shielded.)

ex. `are_shielded_nodes_enabled`

validator/gke_enable_shielded_nodes_test.rego
```diff
@@ -0,0 +1,47 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+package templates.gcp.GCPGKEEnableShieldedNodesConstraintV1
+
+import data.validator.gcp.lib as lib
+
+all_violations[violation] {
```
morgante: Do you mind updating the tests to use the new util we just added (should make them simpler)? #289

mmontan: Nice predicate. Much easier to write tests now!

validator/gke_enable_shielded_nodes.rego
mmontan: Changed all names to indicate that they are negatives.

PR #288, opened 2020-03-06
Reduce boilerplate from test fixtures

morgante:
Use list comprehensions in tests to make them more succinct.

### DIFF
--- a/validator/allowed_resource_types_test.rego
+++ b/validator/allowed_resource_types_test.rego
@@ -26,9 +26,12 @@ import data.test.fixtures.allowed_resource_types.constraints.basic.blacklist as 
 import data.test.fixtures.allowed_resource_types.constraints.basic.whitelist as fixture_constraint_basic_whitelist
 
 test_allowed_resource_types_whitelist_violations {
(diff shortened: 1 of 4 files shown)


morgante:
@briantkennedy I only cleaned up a few, but assuming this pattern looks good to you I'll role it out to the other tests. Should reduce repetition in tests substantially and make them more readable.

validator/allowed_resource_types_test.rego
```diff
@@ -26,9 +26,12 @@ import data.test.fixtures.allowed_resource_types.constraints.basic.blacklist as
 import data.test.fixtures.allowed_resource_types.constraints.basic.whitelist as fixture_constraint_basic_whitelist
 
 test_allowed_resource_types_whitelist_violations {
-	violations := [violation | violations := deny with input.asset as fixture_assets[_]
-		with input.constraint as fixture_constraint_basic_whitelist
-		violation := violations[_]]
+	violations := [violation |
```
briantkennedy: Not saying to do this now, but if this is a common pattern, we should see if there's an easy way to create a rule where the user only has to specify fixture_assets, the input, and violation count.

PR #295, opened 2020-03-07
Constraint to control GKE Master exposure

mmontan:
The constraint ensures that GKE masters is not exposed to unauthorized networks. GKE private endpoints, and GKE public endpoints with master authorized networks that are limited to given subnets are ok.


mmontan:
@morgante this is somewhat related to the GKEEnableAuthorizedNetworks. I didn't add anything to it as the check is CIS-specific and only refer to authorized networks being enabled. 

Do you prefer this more complex logic to be integrated with the other constraint instead?

morgante:
> Do you prefer this more complex logic to be integrated with the other constraint instead?

Yes, I think we should:
1. Add the option of specifying certain networks to the existing constraint
2. Make the new constraint only focused on the private endpoint check.

mmontan:
@morgante  Ok! I'll update the pull request.

We will have two constraints that say:
1) networks need to be within this list.
2) endpoint needs to be private.

The semantic will be slightly different than the integrated constraint, but I think it'll work well and it will be more clear.

morgante:
Sounds good!

And, to be clear, for "networks need to be within this list" I suggest making it an optional param on the existing constraint (if unspecified, *any* network is allowed so long as master authorized networks are configured).

mmontan:
Don't merge yet.  I need to work more on the defaults. I need a way to specify that authorized networks is enabled, and NO networks are added. it will be useful when used in conjuction with private endpoint, so that we can enforce that a master is accessible only from within the cluster.

mmontan:
Ready for review now. Thank you!

Issue #267, opened 2020-02-18
[FORSETI] IAM Policy scanner functionality


hshin-g:
Related to Forseti Security Issue [#3624](https://github.com/forseti-security/forseti-security/issues/3624)

### Scanner

IAM Policy Scanner

**Description:** Cloud IAM policies directly grant access on GCP. To ensure only authorized members and permissions are granted in Cloud IAM policies, IAM policy scanner supports the following:
- Whitelist, blacklist, and required modes.
- Define if the scope of the rule inherits from parents or just self.
- Access to specific organization, billing account, folder, project, bucket and dataset resource types.

### Python Violations Summary
There are three types (modes) of violations that can be generated:

1) **Required** - Generate a violation when there are policy members that do NOT exactly match the members  defined in the “members” field in iam_rules.yaml file. If there are members in the “members” field in iam_rules.yaml file that are missing, generate a violation.

2) **Whitelist** - Generate a violation when there are policy members that do NOT match the members defined in the “members” field in iam_rules.yaml file.

3) **Blacklist** - Generate a violation when there are policy members that DO match the members  defined in the “members ” field in iam_rules.yaml file.

### 1:1 Rego | Python Parity Status
It looks like the Rego rule does do whitelist/blacklist. However, the “required” mode is not covered, which is where the bindings for a rule MUST be included in the resource’s bindings, or a violation will be reported. Additionally, inheritance is not covered in the Rego rule, where the bindings for the children of a resource will also be evaluated, and the bindings of a child resource will be evaluated with its parent's bindings.

### Rego work to do

Create or edit existing Rego [rule](https://github.com/forseti-security/policy-library/blob/master/policies/templates/gcp_iam_allowed_bindings_v1.yaml) to do the following:

1) Allow for ""required"" mode --> currently rule does whitelist/blacklist.

hshin-g:
@morgante : Doable - could probably add required mode as all of the members are included inside the asset.

Inheritance would be difficult to do, and referential (would have to merge policies, etc).

PR #305, opened 2020-03-19
IAM Policy Required Bindings Constraint v1

hshin-g:
### IAM Policy Required Bindings constraint for #267 

This PR creates a IAM Policy Required Bindings constraint that checks for the bindings of IAM policies and checks whether all of the user defined required members are included. If not, generate a violation.

A sample playground for this constraint is provided [here](https://play.openpolicyagent.org/p/707GzQieZ1).

1. **asset_type**
- Restrict which asset type this policy applies to.
  - E.g. cloudresourcemanager.googleapis.com/Project would restrict to only Project resources
2. **role**
- The role to restrict bindings for. Wildcards (*) supported.
  - E.g. roles/owner would only look at bindings for the owner role.
  - E.g. roles/* would look at all roles available.
3. **members**
- Members that are required for the given role. This can be specific users, groups, or domains. Wildcards (*) supported.
  - E.g. 'user:*@google.com' would generate violations if there were no users with domain @google.com.
  - E.g. 'group:*@google.com' would generate violations if there were no groups with domain @google.com.
  - E.g. 'user:myuser111@google.com' would generate violations if user:myuser111@google.com does not exist.

### Also included in this PR:

- Sample constraint
- Tests


hshin-g:
@morgante @briantkennedy Could this get a review from one of you? Thanks!

validator/iam_required_bindings_test.rego
```diff
@@ -0,0 +1,104 @@
...
+# limitations under the License.
+#
+
+package templates.gcp.GCPIAMRequiredBindingsConstraintV1
+
+import data.test.fixtures.iam_required_bindings.assets as fixture_assets
+import data.test.fixtures.iam_required_bindings.constraints as fixture_constraints
+
+# Find all violations on our test cases
+find_violations[violation] {
```
briantkennedy: I think this might be able to leverage some of test_utils.rego

hshin-g: Have changed- PTAL.

briantkennedy: I think you can also remove find_violations

hshin-g: Thanks for the catch! Have fixed.

PR #310, opened 2020-03-20
Add support for testing with multiple rego versions.

### LINKED ISSUE #285: Run rego tests with multiple versions of opa
It looks like we had some issues where a CT was failing on older versions of OPA, we should update our CI tests to run the rego unit tests using all supported versions of opa.

briantkennedy:
fixes #285

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -37,10 +37,6 @@ format: ## Format Rego rules
 .PHONY: build
 build: format build_templates ## Format and build
 
-.PHONY: push_make_image
-push_make_image: ## Construct and push Docker image for Cloud Build CI to gcr.io/config-validator/make
-	@cd cloudbuild && gcloud builds submit --project=config-validator --tag gcr.io/config-validator/make .
-
 .PHONY: check_sample_files
 check_sample_files: ## Make sure each template in policies/templates has one sample file using it in samples/
 	@python3 scripts/check_samples.py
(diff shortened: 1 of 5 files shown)


briantkennedy:
Ready for review, I'll work through the v0.17.0 issues at some point.

Makefile
```diff
@@ -56,3 +52,19 @@ audit: ## Run audit against real CAI dump data
 
 help: ## Prints help for targets with comments
 	@grep -E '^[a-zA-Z._-]+:.*?## .*$$' $(MAKEFILE_LIST) | sort | awk 'BEGIN {FS = ":.*?## "}; {printf "make \033[36m%- 30s\033[0m %s\n", $$1, $$2}'
+
+
+# Cloudbuild doesn't allow us to use the Dockerfile "ARG" feature so we have to
```
morgante: This seems rather hacky. Could we instead add a cloudbuild.yaml in the /cloudbuild/ directory which would invoke the Docker builder and properly pass through arguments? Like this example here - https://cloud.google.com/cloud-build/docs/configuring-builds/substitute-variable-values#using_default_substitutions

briantkennedy: Unfortunately, it doesn't support substitutions for dockerfiles (as far as I can tell).

morgante: Specifically, I'd suggest we leave arguments in the Dockerfile then add a cloudbuild.yaml which does something like this:

````
# Uses the docker build step to build an image called my-image
- name: 'gcr.io/cloud-builders/docker'
  args: ['build', '-t', 'gcr.io/$PROJECT_ID/my-image:$_REGO_VERSION', '., '--build-arg', 'REGO_VERSION=$_REGO_VERSION]
```

briantkennedy: I attempted to convert the change to this format.  Cloudbuild local worked, but when sending to GCP, it wasn't working, so I think this will have to be a future cleanup.

morgante: Fair enough, I think that's reasonable.

PR #311, opened 2020-03-25
Improve make_format error message.

briantkennedy:
Adjust output to print what was wrong and resolution.

### DIFF
--- a/scripts/check_format.sh
+++ b/scripts/check_format.sh
@@ -1,0 +1,15 @@
+#!/bin/bash
+
+set -euo pipefail
+
+diff="$(opa fmt -d lib/ validator/)"
+
+if [[ "$diff" == "" ]]; then
+  exit
+fi
+
+echo "Formating error:"
+echo "$diff"
+echo ""
+echo "Run \"make format\" in your client to fix."
+exit 1
--- a/Makefile
+++ b/Makefile
@@ -43,7 +43,7 @@ check_sample_files: ## Make sure each template in policies/templates has one sam
 
 .PHONY: check_format
 check_format: ## Check that files have been formatted using opa fmt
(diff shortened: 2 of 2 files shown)


scripts/check_format.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
+
+set -euo pipefail
+
+diff="$(opa fmt -d lib/ validator/)"
+
+if [[ "$diff" == "" ]]; then
```
AdrienWalkowiak: I would rather check the return code for the previous command ($? -ne 0)

scripts/check_format.sh
```diff
@@ -0,0 +1,15 @@
...
+  exit
+fi
+
+echo "Formating error:"
+echo "$diff"
+echo ""
+echo "Run \"make format\" to fix."
```
AdrienWalkowiak: nit: suggest saying run make format locally to reproduce this error

scripts/check_format.sh
briantkennedy: it exits 0 diff or no diff.

scripts/check_format.sh
briantkennedy: fixed

scripts/check_format.sh
AdrienWalkowiak: I believe you can redirect the error to the standard output:

diff="$(echo opa fmt -d lib/ validator/ 2>&1)"

if [[ $? -ne 0 ]];
then

briantkennedy: If there is a diff, it exits 0.  If there is no diff, it exits 0.  We can't rely on the exit code for whether or not there is a diff.

AdrienWalkowiak: sounds good. it's too bad opa does not behave in a typical way

PR #319, opened 2020-03-26
Update makefile to add common dev tasks using opa docker image.

### LINKED ISSUE #318: Add OPA test and format commands to makefile using Docker image
Adding these common commands to the makefile would improve the dev experience. This allows devs to test the policies regardless of platform and not have to download the various versions of OPA.

gkowalski-google:
Resolves #318

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -15,10 +15,15 @@
 # Make will use bash instead of sh
 SHELL := /usr/bin/env bash
 
-OPA_IMAGE := opa
-OPA_IMAGE_TAG_V15 := 0.15.1
-OPA_IMAGE_TAG_V16 := 0.16.2
-OPA_IMAGE_URL := openpolicyagent
+CV_IMAGE_URL := gcr.io/config-validator
+
+POLICY_TOOL_IMAGE := policy-tool
+POLICY_TOOL_IMAGE_TAG := latest
+
+REGO_IMAGE_V15 := rego-v0.15.0
+REGO_IMAGE_V16 := rego-v0.16.0
+REGO_IMAGE_TAG := latest
+
 
 # The .PHONY directive tells make that this isn't a real target and so
 # the presence of a file with that name won't cause this target to stop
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -15,13 +15,24 @@
 # Make will use bash instead of sh
 SHELL := /usr/bin/env bash
 
+OPA_IMAGE := opa
```
morgante: Should we use the image we maintain? https://github.com/forseti-security/policy-library/blob/master/cloudbuild.yaml

gkowalski-google: Let me try using that one. With the official OPA image, `fmt -d` does not work because `diff` cannot be found. Maybe it will work with this image.

gkowalski-google: I'm not running it right.
```
docker run -it --rm \
                -v $(CURDIR):/workspace \
                gcr.io/config-validator/rego-v0.16.0:latest \
                test
```
produces error:
```
make: *** No rule to make target 'test'.  Stop.
make: *** [docker_test] Error 2
```

Makefile
```diff
@@ -15,13 +15,24 @@
...
 # The .PHONY directive tells make that this isn't a real target and so
 # the presence of a file with that name won't cause this target to stop
 .PHONY: test
 test: ## Test constraint templates via OPA
 	@echo "Running OPA tests..."
 	@opa test --timeout 30s -v lib/ validator/ --explain=notes
 
+.PHONY: docker_test
+docker_test: ## Run tests using OPA docker image
+	docker run -it --rm \
```
briantkennedy: can you adjust this to run tests using opa 0.15.x and .016.2

Makefile
```diff
@@ -45,6 +56,13 @@ check_sample_files: ## Make sure each template in policies/templates has one sam
 check_format: ## Check that files have been formatted using opa fmt
 	@./scripts/check_format.sh
 
+.PHONY: docker_check_format
+docker_check_format: ## Check format of rego using OPA docker image
+	docker run -it --rm \
```
briantkennedy: Let's keep this consistent with the cloudbuild and use docker 0.15.x.

Makefile
gkowalski-google: Updated using OPA image to run tests on 0.15.1 + 0.16.2 by running `make docker_test`.

Makefile
gkowalski-google: Updated.

Makefile
morgante: This command worked for me, you need to set the working directory (or we could update the Dockerfile to include WORKDIR):
```
docker run -it --rm \
  -v "$(pwd):/workspace" \
  -w=/workspace \
  gcr.io/config-validator/rego-v0.16.0:latest \
  test
```

gkowalski-google: Nice, thanks! This works for me now.

gkowalski-google:
> Alternatively, we could look at using the opa image in cloud build and using cloud-build-local to execute things at pre-submit time

@morgante @briantkennedy Let me know what your preference is. I was using the OPA image because I couldn't get the CV image working, but Morgante helped with that. I'm happy to make the changes. The motivation for me was to make it easy to run the tests on mac, where I frequently run into issues with certain software blocking me from running the OPA executable.

morgante:
> @morgante @briantkennedy Let me know what your preference is. I was using the OPA image because I couldn't get the CV image working, but Morgante helped with that. I'm happy to make the changes. The motivation for me was to make it easy to run the tests on mac, where I frequently run into issues with certain software blocking me from running the OPA executable.

I think we should use the same image in both. It'd probably be easiest for you to update the Makefile to use the CV image.

gkowalski-google:
I've updated to run the tests using the CV rego image. Also added in targets to run the lint tests, check sample files, and check the formatting. Everything working for me locally.

Issue #266, opened 2020-02-18
[FORSETI] Firewall Rules scanner functionality


hshin-g:
Related to Forseti Security Issue [#3624](https://github.com/forseti-security/forseti-security/issues/3624)

### Scanner

Firewall Rules Scanner

**Description:** Network firewall rules protect your network & organization by only allowing desired traffic into and out of your network. The firewall rules scanner can ensure that all your network’s firewalls are properly configured.

### Python Violations Summary
There are four types (modes) of violations that can be generated:

1) **Matches** - Find the Firewall rules that don't match the required policy found under the “match_policies” field in firewall_rules.yaml file.

   - If a rule in the “match_policies” field in firewall_rules.yaml does NOT exist in the Firewall rules, generate a violation and a recommended action to INSERT the missing rule into the Firewall rules (in order to match the “match_policies” that have been defined). This is referential as it has to look at ALL the Firewall rules to see whether the rule in "match_policies" exist- i.e. is a match. Only when it sees that the rule in "match_policies" does NOT exist in the Firewall rules, does it generate a violation.

   - If a Firewall rule does NOT exist in the rules found under the “match_policies” field in the firewall_rules.yaml file, generate a violation and a recommended action to DELETE the Firewall rule (in order to match the “match_policies” that have been defined). This seems to be a whitelist function, as it just needs to check whether a Firewall rule exists in a user-defined list of "match_policies". If it does NOT exist, generate a violation.

2) **Required** - Find the missing rules that are required from the “match_policies” field in firewall_rules.yaml file.

   - If a rule in the “match_policies” field in firewall_rules.yaml does NOT exist in the Firewall rules, generate a violation and a recommended action to INSERT the missing match_rule into the Firewall rules (in order to match the “match_policies” that have been defined). This is referential as it is required that ALL the Firewall rules be evaluated to see whether the required rule in "match_policies" is included.

3) **Whitelist** - Find Firewall rules that do NOT match the whitelisted rules in firewall_rules.yaml file.

4) **Blacklist** - Find the blacklisted policies and return violations if there are any Firewall rules that are blacklisted in firewall_rules.yaml file.

### 1:1 Rego | Python Parity Status
The Firewall Rego rule generates violations by listing rules to match - i.e. if a firewall rule matches the parameters of the Rego rules, it will generate violations. It does not look like WHITELIST, REQUIRED, or MATCHES modes are covered. BLACKLIST seems to be covered here. 

### Rego work to do

[...]

PR #322, opened 2020-04-01
Firewall Rule Allowlist Constraint v1

hshin-g:
### Firewall Rule Allowlist constraint for #266 

This PR creates a Firewall Rule Allowlist constraint that restricts firewall rules to only those that exactly match the user defined parameters. If not, generate a violation.

This follows similar logic to that expressed in the [gcp_restricted_firewall_rules_v1.yaml](https://github.com/forseti-security/policy-library/blob/master/policies/templates/gcp_restricted_firewall_rules_v1.yaml) template.

A sample playground for this constraint is provided [here](https://play.openpolicyagent.org/p/urvhARcu4R).

1. **rules**
- Generate violations if resource data does not match the list of rules.
2. **direction**
- Direction of the rules to restrict.
  - "INGRESS", "EGRESS", "any"
3. **rule_type**
- Type of the rules to restrict.
  - "allowed", "denied", "any"
4. **port**
- Port of the rules to restrict. This parameter is tighly coupled with the protocol paramter: the rule will be marked as in violation if the parameter protocol:port pair does NOT match one of its protocol:port listed.
  - Valid values are:
      * single number
      * ranges (e.g 22-25) - Note: generate a violation if FW rule has a range that is greater than or less than the passed range.
      * 'any' - when set to any, the ports of the rule will NOT be a factor to determine if the FW rule is restricted. Default value.
       * 'all' (the rule matches all ports for a given protocol)
5. **protocol**
- Protocol of the rules to restrict. This parameter is tighly coupled with the port paramter: the rule will be marked as in violation if the parameter protocol:port pair does NOT match one of its protocol:port listed, or if protocol does not match.
    - Valid values are:
        * single protocol (e.g 'tcp', 'icmp' etc.)
        * 'any' - when set to any, the protocol of the rule will NOT be a factor to determine if the FW rule is restricted. Default value.
         * 'all' (if the rule matches all protocols)
6. **source_ranges**
- Source ranges of the rules to restrict. A fw rule will generate a violation if NONE of its source CIDR ranges is an exact match of ANY of the CIDR blocks passed in this parameter.
    - Valid values are:
       * 'X.X.X.X/XX' - i.e valid CIDR blocks
       * 'any' - when set to any, the source ranges of the rule will NOT be a factor to determine if the FW rule is restricted. Default value.
       * '*' - all CIDR blocks match: this only checks that a source range exists in the restricted FW rule
7. **source_tags**
- Source tags of the rules to restrict. A fw rule will generate a violation if NONE of its source tags is a match of ANY of the tag patterns (regex) passed in this parameter.
    - Valid values are:
       * 'mytag-*' - valid regex patterns for source tags
       * 'any' - when set to any, the source tags of the rule will NOT be a factor to determine if the FW rule is restricted. Default value.
        * '*' - all source tags match: this only checks that a source tag exists in the restricted FW rule
8. **source_service_accounts*
- Source service accounts of the rules to restrict. A fw rule will generate a violation if NONE of its source service accounts is a match of ANY of the service account patterns (regex) passed in this parameter.
    - Valid values are:
       * 'mytag-*' - valid regex patterns for source tags
        * 'any' - when set to any, the source tags of the rule will NOT be a factor to determine if the FW rule is restricted. Default value.
        * '*' - all source service accounts match: this only checks that a source service account exists in the restricted FW rule
9. **target_ranges**
- Target (or destination) ranges of the rules to restrict. A fw rule will generate a violation if NONE of its destination CIDR ranges is an exact match of ANY of the CIDR blocks passed in this parameter.
    -  Valid values are:
         * 'X.X.X.X/XX' - i.e valid CIDR blocks
         * 'any' - when set to any, the target ranges of the rule will NOT be a factor to determine if the FW rule is restricted. Default value.
         * '*' - all CIDR blocks match: this only checks that a target / destination range exists in the restricted FW rule
10. **target_tags**
- Target tags of the rules to restrict. A fw rule will generate a violation if NONE of its target tags is a match of ANY of the tag patterns (regex) passed in this parameter.
    -Valid values are:
      * 'mytag-*' - valid regex patterns for target tags
      * 'any' - when set to any, the target tags of the rule will NOT be a factor to determine if the FW rule is restricted. Default value.
      * '*' - all target tags match: this only checks that a target tag range exists in the restricted FW rule
11. **target_service_accounts**
- Target service accounts of the rules to restrict. A fw rule will generate a violation if NONE of its target service accounts is a match of ANY of the service account patterns (regex) passed in this parameter.
   - Valid values are:
      * 'mytag-*' - valid regex patterns for target tags
      * 'any' - when set to any, the target tags of the rule will NOT be a factor to determine if the FW rule is restricted. Default value.
      * '*' - all target service accounts match: this only checks that a target service account exists in the restricted FW rule

### Also included in this PR:

- Sample constraint
- Tests


morgante:
Why does this need to be a new constraint template?

Especially given the `mode` parameter, it seems like this should simply be a modification of the existing constraint. Can you explain more about the key differences?

hshin-g:
I tried to utilize the existing constraint template but I wasn't able to get the allowlist functionality to work without adding a lot more code to it, and it wasn't super optimal. I created this new constraint so it was cleaner on what was being looked for, and wanted to also add more transparency in what was actually generating violations (e.g. source ranges don't match vs. just a generic violation).

morgante:
> I tried to utilize the existing constraint template but I wasn't able to get the allowlist functionality to work without adding a lot more code to it, and it wasn't super optimal.

I'm surprised it added that much code, but maybe I'm misunderstanding this new constraint.

Isn't this new constraint simply the inverse of the denylist version? ie. any rule *not* matching this constraint will trigger a violation?

hshin-g:
You're right, Morgante- I must have done something wrong in the beginning to lead me down this route. Closing this PR in lieu of #326 - thanks!

PR #331, opened 2020-04-10
IAM Custom Role Permissions Constraint V1

hshin-g:
### IAM Custom Role Permissions Constraint V1

This PR creates an IAM Policy Custom Role Permissions constraint that checks whether the user defined permissions are either allowed or denied in a user defined custom role.

A sample playground for this constraint is provided [here](https://play.openpolicyagent.org/p/U4DDvP39H1).

1. **mode**
- Enforcement mode, [denylist, allowlist] , defaults to allowlist.
2. **title**
- The custom role title to scan for, found on the IAM Roles page.
    - E.g. 'My Custom Role for BigQuery'
3. **permissions**
- Permissions to either allow or deny for the given custom role, depending on mode; Wildcards (*) supported.
    - E.g. 'bigquery.datasets.*' will search for any permissions that start with bigquery.datasets, such as bigquery.datasets.get
    - E.g. 'bigquery.**' will search for any permissions that start with bigquery, such as bigquery.savedqueries.get, bigquery.config.get
    - E.g. '*' will search for any permissions

### Also included in this PR:

- Sample constraint
- Tests
- Added a function to `test_utils` to test for a field other than `resource`.

### DIFF
--- a/samples/iam_custom_role_permissions.yaml
+++ b/samples/iam_custom_role_permissions.yaml
@@ -1,0 +1,28 @@
+# Copyright 2020 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+apiVersion: constraints.gatekeeper.sh/v1alpha1
+kind: GCPIAMCustomRolePermissionsConstraintV1
+metadata:
+  name: allowlist_custom_role_permissions
+  annotations:
+    description: Custom BigQuery role must only have specific permissions
+spec:
+  severity: high
+  parameters:
+    mode: allowlist
+    title: "My Custom BigQuery Role"
+    permissions:
+    - "bigquery.datasets.get"
+    - "bigquery.tables.*"
--- a/samples/iam_custom_role_permissions.yaml
+++ b/samples/iam_custom_role_permissions.yaml
@@ -15,7 +15,7 @@
 apiVersion: constraints.gatekeeper.sh/v1alpha1
 kind: GCPIAMCustomRolePermissionsConstraintV1
 metadata:
-  name: allowlist_custom_role_permissions
+  name: allowlist-custom-role-permissions
   annotations:
     description: Custom BigQuery role must only have specific permissions
 spec:
--- a/validator/iam_custom_role_permissions.rego
+++ b/validator/iam_custom_role_permissions.rego
@@ -30,7 +30,10 @@ deny[{
 
 	asset_permissions := asset.resource.data.includedPermissions[_]
 	asset_title := asset.resource.data.title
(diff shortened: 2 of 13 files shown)


morgante:
@hshin-g Looks like tests are failing.

hshin-g:
Thanks for the check! It looks like my branch hadn't been updated with the newest master. _Should_ be fixed now- thanks!

samples/iam_custom_role_permissions.yaml
```diff
@@ -0,0 +1,28 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+apiVersion: constraints.gatekeeper.sh/v1alpha1
+kind: GCPIAMCustomRolePermissionsConstraintV1
+metadata:
+  name: allowlist_custom_role_permissions
```
morgante, line 18:
```suggestion
  name: allowlist-custom-role-permissions
```

validator/iam_custom_role_permissions.rego
```diff
@@ -0,0 +1,71 @@
...
+}] {
+	constraint := input.constraint
+	lib.get_constraint_params(constraint, params)
+	asset := input.asset
+
+	asset.asset_type == "iam.googleapis.com/Role"
+
+	asset_permissions := asset.resource.data.includedPermissions[_]
+	asset_title := asset.resource.data.title
+	lower(asset_title) == lower(params.title)
```
morgante, line 33: Should we make this check optional? It seems like having to choose a specific role title might be messy. In many scenarios we might want to block a permission from all roles.

Maybe make this a regex w/ a default of matching everything?

validator/test_utils.rego
```diff
@@ -40,6 +40,13 @@ check_test_violations_resources(test_assets, test_constraints, test_template, ex
 	resource_names == expected_resource_names
 }
 
+# This is to check for other field names from violations besides resource
+check_test_violations_field_names(test_assets, test_constraints, test_template, field_name, expected_resource_names) {
```
morgante, line 44:
```suggestion
check_test_violations_metadata(test_assets, test_constraints, test_template, field_name, field_values) {
```

samples/iam_custom_role_permissions.yaml
hshin-g: Fixed!

validator/iam_custom_role_permissions.rego
hshin-g: That's a good point. I created a default of "*" if title is not defined, so it will just look at all titles.

validator/test_utils.rego
hshin-g: Fixed!

validator/iam_custom_role_permissions.rego
hshin-g: Appreciate the quick review!

validator/iam_custom_role_permissions.rego
```diff
@@ -0,0 +1,83 @@
...
+target_match_count(mode) = 1 {
+	mode == "allowlist"
+}
+
+check_asset_title(asset_title, params_title) {
+	params_title == "*"
+}
+
+check_asset_title(asset_title, params_title) {
+	params_title != "*"
```
morgante, line 71: FYI it's probably better to allow Regex for fields like this than to have an arbitrary "*" magic key.

Issue #329, opened 2020-04-10
Request to allow whitelist buckets for GCPStorageBucketWorldReadableConstraintV1


aimjwizards:
Please add in exemption feature in constraint in order to exempt/exclude specific buckets in the [GCPStorageBucketWorldReadableConstraintV1](https://github.com/forseti-security/policy-library/blob/master/policies/templates/gcp_storage_bucket_world_readable_v1.yaml) template 

Certain buckets such as CDN buckets are ok to be public and should be allowed to be specified as exemptions  like [example](https://github.com/forseti-security/policy-library/blob/fed9cc42d71896fc9bccf63d05d3a7b8902e6dbb/policies/templates/gcp_storage_bucket_retention_v1.yaml#L58).

This allows a bipartition of buckets into "ok to public" and "not ok to public" which is useful for many users

PR #334, opened 2020-04-13
Add Exemptions to GCP Storage Bucket World Readable V1 Constraint

hshin-g:
Resolves Issue #329.

This PR adds functionality to exempt (skip over) specific GCS buckets from being scanned in the GCP Storage Bucket World Readable V1 Constraint.

A sample playground can be found [here](https://play.openpolicyagent.org/p/J7r8vUEINa).


briantkennedy:
Thanks for taking care of this!

PR #348, opened 2020-05-08
Add kpt setter for constraint target

### LINKED ISSUE #336: Add ability to specify some parameters to policy bundle
## Story
Users will need to specify some parameters to the policy bundle. For example, users will need to provide the target for the constraints. Most default to target the org, but some users run Forseti against a folder and the constraints would need to be updated.

## Proposed Solution
Explore using kpt setters or kustomize.

gkowalski-google:
Create a kpt setter to set the constraint target. Add annotations to the samples.

Resolves #336


samples/appengine_location.yaml
```diff
@@ -5,7 +5,7 @@ metadata:
 spec:
   severity: high
   match:
-    target: ["organization/*"]
+    target: ["organization/*"] # {"$ref":"#/definitions/io.k8s.cli.setters.target"}
```
morgante, line 8: Does this correctly still yield an array when replaced?

gkowalski-google: Yes, running the setter like `kpt cfg set ./policies/constraints target "folder/*"` does not change the format. However, I just noticed that running the policy bundle fn removes the kpt annotations. Is that expected? I was planning on getting the bundle into the constraints dir and then running this, but it won't work when the annotations are removed by the fn.

morgante: No that's not intentional.

gkowalski-google: Before merging this in, I want to make sure this will work properly against the output of the policy bundle function. I have found that running the policy bundle kpt fn removes all comments/annotations, so running this setter after getting the bundle will not work. I have created an [issue on the kpt repo](https://github.com/GoogleContainerTools/kpt/issues/628) to find out more.

gkowalski-google: @morgante This PR has been updated and confirmed working with the latest version of kpt, I think we are good to merge this in, would you mind reviewing once more?

PR #351, opened 2020-05-19
Add ancestry path metadata to all templates

katze120:
This is for converging with CSPM templates, and also to make locating the violating resource easier in general.


morgante:
Instead of adding these to all constraint templates, maybe we should add it directly to Config Validator. @briantkennedy thoughts?

If this is specific for CFT Scorecard, we could also inject it there.

briantkennedy:
Yeah, I agree with @morgante, this would be a pretty quick change in FCV and we wouldn't have to ensure that all templates have it.

briantkennedy:
I was about to file an issue for FCV when I took a look.  This should already be available in the FCV violation at `.resource.ancestry_path`.

morgante:
@katze120 Does the above work for you or do we need to do some additional work in Scorecard to expose it?

katze120:
I took a closer look at the code and actual Violation output and don't think it's already there.. raised https://github.com/forseti-security/config-validator/issues/140

This will need to be followed up with changes in scorecard and downstream integrations as location of this info will change in scorecard output.

PR #360, opened 2020-06-05
Add test to confirm make build was run.

### LINKED ISSUE #359: Prevent unsynced policies
To prevent a recurrence of #358 (which took way too much time to debug for a trivial fix), we should add CI checks to confirm that `make build` has been run.

morgante:
Fixes #359


scripts/check_build.sh
```diff
@@ -0,0 +1,45 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -euo pipefail
+
+# Confirm we have run `make build` by:
+# 1) Running make build in a tmp directory
```
briantkennedy, line 20: I think we can leverage git and run "make build" then "git status --short" to see if files are changed (it will print output if changes exist, nothing if no changes).

morgante: That would work fine in CI, but I also want this to be runnable locally (where we don't want to run `make build` on the live git install).

briantkennedy: I'm not quite sure I understand the local case, what's the user/developer flow look like there?

morgante: For things like git precommit checks it's helpful to be able to run simple linters like this locally.

It's admittedly not an especially common case, but I don't think it's worth changing to something else given this.

briantkennedy:
Sorry for the delay, things got kind of hectic this past week

Issue #362, opened 2020-06-07
Change to more inclusive language


morgante:
The terms "blacklist" and "whitelist" are [unrecommended](https://developers.google.com/style/inclusive-documentation#features-and-users).

We currently have [55 instances](https://github.com/forseti-security/policy-library/search?q=blacklist&unscoped_q=blacklist) of `blacklist` and [68 instances](https://github.com/forseti-security/policy-library/search?q=whitelist) of `whitelist` which we should consider changing. In most cases, `denylist` and `allowlist` are descriptive alternatives.

Let's also make sure that new constraints use inclusive language and ideally have a reference in developer documentation.

briantkennedy:
Adding the type to the field name gets a bit redundant while reading the field, let's go for `deny` and `allow`.

PR #363, opened 2020-06-07
Change iam allowed bindings constraint to use allowlist/denylist.

morgante:
Ref #362


samples/iam_block_public.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: constraints.gatekeeper.sh/v1alpha1
-kind: GCPIAMAllowedBindingsConstraintV1
+kind: GCPIAMAllowedBindingsConstraintV2
 metadata:
-  name: blacklist_all_users
+  name: block_allusers
```
briantkennedy, line 4: for consistency `deny_allusers`?

samples/iam_block_public.yaml
```diff
@@ -12,7 +12,7 @@ spec:
     - "organization/*"
     exclude: [] # optional, default is no exclusions
   parameters:
-    mode: blacklist
+    mode: denylist
```
briantkennedy, line 15: I'm a fan of succinctness, any thoughts on `deny` over having the type in the name?

morgante:
> I like the direction. I think we may want to look at how we can migrate our users to the new formats (is this something we can automate across all templates or per template, or if it will be by hand for each template), and if there's any way to automate making this adjustment on each template itself.

I'm not overly worried about a need for additional automation here. Making the changes "by hand" with grep was pretty fast so I think that approach can be used for additional ones.

As for end-users, we'll retain the old templates in `legacy/` so it's just a matter of updating to v2 when they're comfortable. Unless we hear customer questions/concerns, I'm inclined to think that's sufficient.

samples/iam_block_public.yaml
morgante: We already use [denylist](https://github.com/forseti-security/policy-library/search?q=denylist&unscoped_q=denylist) on quite a few templates, so I'm inclined to keep that as the standard unless you think there's a strong reason to change.

briantkennedy: No strong reason.  If allowlist/denylist is already the def-facto standard let's keep using that.

PR #368, opened 2020-06-16
minor fix in iam audit log tests

xingao267:
The name of the resource in the `data.json` is `//cloudresourcemanager.googleapis.com/ignore-asset-type`

### DIFF
--- a/validator/iam_audit_log_test.rego
+++ b/validator/iam_audit_log_test.rego
@@ -15,7 +15,7 @@ test_audit_log_skip_good_project {
 }
 
 test_audit_log_skip_irrelevant_asset_type {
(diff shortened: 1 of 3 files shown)


validator/iam_audit_log_test.rego
```diff
@@ -15,7 +15,7 @@ test_audit_log_skip_good_project {
 }
 
 test_audit_log_skip_irrelevant_asset_type {
-	not resources_in_violation["//cloudresourcemanager.googleapis.com/projects/ignore-asset-type"]
+	not resources_in_violation["//cloudresourcemanager.googleapis.com/ignore-asset-type"]
```
morgante, line 18: For consistency, we should fix this in the data fixture.

xingao267: Done

PR #371, opened 2020-06-25
Update docs generator function to accept flexible config.

morgante:
This allows for generating alternative docs formats.


morgante:
@gkowalski-google PTAL

Issue #375, opened 2020-09-11
make generate_docs generates `EACCES: permission denied` error


xingao267:
At commit ccab976557943550d764659abce52372ecaba94d

```
$ make generate_docs 
Generating docs with kpt...
Error: EACCES: permission denied, stat '/docs/func.yaml'
make: *** [Makefile:147: generate_docs] Error 2
```

gkowalski-google:
@xingao267 Can you take a look at what the permissions are for `/docs/func.yaml`?

xingao267:
I do have write permission on the file.
```
$ ls -la docs/func.yaml 
-rw-rw-rw- 1 xingao primarygroup 1054 Sep 11 15:35 docs/func.yaml
```

It seems the permission issue is from within the container.

xingao267:
The cause is the generate_docs container sets the default user to be [`node`](https://github.com/forseti-security/policy-library/blob/master/bundler/build/generate_docs.Dockerfile#L27), which does not have `write` permission to the files under `docs` dir by default when those files are mounted into the container. I bypassed it by adding `-u root` in the `docker run` command in the `make generate_docs` command.

xingao267:
I'm still getting this error. The generate doc dev container needs to be rebuilt and published i guess.

gkowalski-google:
The updated docker image was successfully built and published on Friday.

PR #377, opened 2020-09-18
fix: Update build image to use root

morgante:
Fixes https://github.com/forseti-security/policy-library/issues/375


gkowalski-google (approved):
LGTM

PR #378, opened 2020-09-18
add healthcare best practices v1 bundle

xingao267:
The selection of policies are based on https://github.com/GoogleCloudPlatform/healthcare-data-protection-suite/tree/master/templates/policygen/forseti/overall


xingao267:
+R: @umairidris

umairidris:
Do we want a dedicated 'healthcare' bundle? Is this bundle more restrictive than others? If so, maybe it can be named something more general such as 'strict' or something? If these configs are those that most non-healthcare customers won't use then naming it healthcare makes sense to me.

xingao267:
> Do we want a dedicated 'healthcare' bundle? Is this bundle more restrictive than others? If so, maybe it can be named something more general such as 'strict' or something? If these configs are those that most non-healthcare customers won't use then naming it healthcare makes sense to me.

I don't think it's more restrictive than other [bundles](https://github.com/forseti-security/policy-library/blob/master/docs/index.md#policy-bundles), but just according to different use cases/benchmarks/purposes, e.g. CIS, GKE hardening or healthcare. This new bundle might also apply to other regulated industries (financial, etc), but because I don't have any experience with those verticals, so don't want to claim it to be that general.

umairidris:
Sounds good. If Morgante is ok with it this LGTM.

Just out of curiosity, can a user install multiple bundles? Does it aggregate them together?

xingao267:
> Sounds good. If Morgante is ok with it this LGTM.
> 
> Just out of curiosity, can a user install multiple bundles? Does it aggregate them together?
It looks like they can be aggregated together. I tried installing both [cis-v1.0](https://github.com/forseti-security/policy-library/blob/master/docs/bundles/cis-v1.0.md) and [gke-hardening-v2019.11.11](https://github.com/forseti-security/policy-library/blob/master/docs/bundles/gke-hardening-v2019.11.11.md) and they are union'ed.

morgante:
> This new bundle might also apply to other regulated industries (financial, etc), but because I don't have any experience with those verticals, so don't want to claim it to be that general.

Yeah, keeping this to healthcare makes sense. We can always add additional bundles if desired.

morgante (commented):
Should we simply call it "healthcare-v1" - also does this align to HIPAA or HITRUST?

xingao267:
> Should we simply call it "healthcare-v1" - also does this align to HIPAA or HITRUST?

healthcare-v1 sgtm. And yes, these policies help customers align with HIPAA, but i'm not sure if including HIPAA in the naming would give users the impression that these policies are all that they need to be HIPAA compliant, which we don't guarantee. @rtal wdyt?

rtal:
That makes sense, I think leaving out the compliance framework names SGTM. Would healthcare-baseline-v1 or something similar make sense, just in case we need to have other healthcare-x-vN templates?

gkowalski-google (approved):
LGTM

xingao267:
healthcare-baseline-v1 SGTM. Changed.

morgante (approved):

PR #390, opened 2020-12-23
Add check for uniform bucket access level

### LINKED ISSUE #388: The policy "storage bucket policy only" is using deprecated argument bucketPolicyOnly
`gcp_storage_bucket_policy_only_v1.yaml` and `storage_bucket_policy_only.rego` are using only the deprecated argument `bucketPolicyOnly` in the validation:

https://github.com/forseti-security/policy-library/blob/73fdd2fee8f9aa71f66e6cc1d1c781a712f6cf73/policies/templates/gcp_storage_bucket_policy_only_v1.yaml#L79

https://github.com/forseti-security/policy-library/blob/73fdd2fee8f9aa71f66e6cc1d1c781a712f6cf73/validator/storage_bucket_policy_only.rego#L41

They should be checking for both `bucketPolicyOnly`  and  `UniformBucketLevelAccess` in this validation until `bucketPolicyOnly` is removed in the next major release of the provider.

daniel-cit:
This pull requests updates `validator/storage_bucket_policy_only.rego` and derived/related files to consider the `uniformBucketLevelAccess` property in addition of the deprecated `bucketPolicyOnly`.

Closes #388

### DIFF
--- a/policies/templates/gcp_storage_bucket_policy_only_v1.yaml
+++ b/policies/templates/gcp_storage_bucket_policy_only_v1.yaml
@@ -65,7 +65,7 @@ spec:
            	asset.asset_type == "storage.googleapis.com/Bucket"
            
            	bucket := asset.resource.data
-           	bucket_policy_enabled(bucket) != true
+           	bucket_policy_enabled(bucket) != true; uniform_bucket_level_access_enabled(bucket) != true
            
            	message := sprintf("%v does not have bucket policy only enabled.", [asset.name])
            	metadata := {"resource": asset.name}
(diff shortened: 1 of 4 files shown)


policies/templates/gcp_storage_bucket_policy_only_v1.yaml
```diff
@@ -66,6 +66,7 @@ spec:
            
            	bucket := asset.resource.data
            	bucket_policy_enabled(bucket) != true
+           	uniform_bucket_level_access_enabled(bucket) != true
```
morgante, line 69: I don't think the logic here is right.

We need _either_ bucket_policy_enabled or uniform_bucket_level_access_enabled not both.

morgante (changes requested):

policies/templates/gcp_storage_bucket_policy_only_v1.yaml
daniel-cit: In the REGO we are trying to find a match for a **violation**, which happens when the property value is not `true`.

the original version of the code, the check was:

```
bucket_policy_enabled(bucket) != true
```
 which rises a **violation** if `bucket_policy_enabled` is not `true`

now the code rises a violation if  `bucket_policy_enabled` is not `true` **AND** `uniform_bucket_level_access_enabled` is not `true`


| bucket_policy_enabled | uniform_bucket_level_access_enabled | result |
|--|--|--|
| true | true | OK |
| true | false | OK |
| false | true | OK |
| false | false | violation |

morgante: You're right, I misunderstood the logic. Thanks, this looks right.

morgante (approved):

PR #410, opened 2021-10-26
CIS 6.2.1 Ensure that the 'log_checkpoints' database flag for Cloud SQL PostgreSQL instance is set to 'on'

palani-ram-google-partner:
CIS 6.2.1 Ensure that the 'log_checkpoints' database flag for Cloud SQL PostgreSQL instance is set to 'on'

### DIFF
--- a/validator/postgresql_log_checkpoints_test.rego
+++ b/validator/postgresql_log_checkpoints_test.rego
@@ -19,20 +19,28 @@ import data.validator.gcp.lib as lib
 import data.validator.test_utils as test_utils
 
 import data.test.fixtures.postgresql_log_checkpoints.assets.no_settings as fixture_no_settings
+import data.test.fixtures.postgresql_log_checkpoints.assets.required as fixture_no_violation
 import data.test.fixtures.postgresql_log_checkpoints.assets.violations as fixture_violation
 
 import data.test.fixtures.postgresql_log_checkpoints.constraints as fixture_constraint
 
 template_name := "GCPPostgreSQLCheckpointsConstraintV1"
 
-#1. postgresql without correct key
-test_postgresql_log_checkpoints_without_correctkey {
+#1. postgresql with correct key
+test_postgresql_log_checkpoints_required_databaseflag {
+	expected_resource_names := {"//cloudsql.googleapis.com/projects/my-test-project/instances/tf-pg-ha-62380f9c-required"}
+	test_utils.check_test_violations_count(fixture_no_violation, [fixture_constraint], template_name, 1)
+	test_utils.check_test_violations_resources(fixture_no_violation, [fixture_constraint], template_name, expected_resource_names)
+}
+
+#2. postgresql without correct key
+test_postgresql_log_checkpoints_violated_databaseflag {
 	expected_resource_names := {"//cloudsql.googleapis.com/projects/my-test-project/instances/tf-pg-ha-62380f9c-violation"}
 	test_utils.check_test_violations_count(fixture_violation, [fixture_constraint], template_name, 1)
 	test_utils.check_test_violations_resources(fixture_violation, [fixture_constraint], template_name, expected_resource_names)
 }
 
-#2. An instance without settings configured at all (settings doesn't exist).
+#3. An instance without settings configured at all (settings doesn't exist).
 test_postgresql_log_checkpoints_no_settings {
 	expected_resource_names := {"//cloudsql.googleapis.com/projects/my-test-project/instances/tf-pg-ha-62380f9c-no-settings"}
 	expected_field_name := "key_in_violation"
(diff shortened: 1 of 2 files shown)


validator/postgresql_log_checkpoints_test.rego
```diff
@@ -0,0 +1,51 @@
...
+import data.test.fixtures.postgresql_log_checkpoints.assets.no_settings as fixture_no_settings
+import data.test.fixtures.postgresql_log_checkpoints.assets.no_violations as fixture_no_violation
+import data.test.fixtures.postgresql_log_checkpoints.assets.violations as fixture_violation
+
+import data.test.fixtures.postgresql_log_checkpoints.constraints as fixture_constraint
+
+template_name := "GCPPostgreSQLCheckpointsConstraintV1"
+
+#1. postgresql with correct key
+test_postgresql_log_checkpoints_no_violations {
```
morgante, line 30: This test / fixture is confusingly named and I'm not sure if the logic is right. If it's meant to have no violations, you should *not* be expecting a violation. Please update.

validator/postgresql_log_checkpoints.rego
```diff
@@ -0,0 +1,48 @@
...
+
+	message := sprintf("On this resource %v Ensure the required key is '%v' and the value is ON.", [asset.name, key])
+	metadata := {"resource": asset.name, "key_in_violation": key}
+}
+
+# check for log_checkpoints under settings databaseflags - no violation
+settings_databaseflags(settings) {
+	setdata := settings.databaseFlags[_]
+	setdata.name == "log_checkpoints"
+	setdata.value != "on"
```
morgante, line 47: Aren't you looking for cases where it *is* on?

morgante (changes requested):

validator/postgresql_log_checkpoints_test.rego
palani-ram-google-partner: @morgante I have fixed the naming for the correct key we are checking.

validator/postgresql_log_checkpoints.rego
palani-ram-google-partner: @morgante I am looking for the case where it is not ON condition.

palani-ram-google-partner:
@morgante Changed the required files and CFT scorecard is working fine too.Please review this PR.

validator/postgresql_log_checkpoints_test.rego
```diff
@@ -0,0 +1,51 @@
...
+import data.test.fixtures.postgresql_log_checkpoints.assets.no_violations as fixture_no_violation
+import data.test.fixtures.postgresql_log_checkpoints.assets.violations as fixture_violation
+
+import data.test.fixtures.postgresql_log_checkpoints.constraints as fixture_constraint
+
+template_name := "GCPPostgreSQLCheckpointsConstraintV1"
+
+#1. postgresql with correct key
+test_postgresql_log_checkpoints_with_correctkey {
+	expected_resource_names := {"//cloudsql.googleapis.com/projects/my-test-project/instances/tf-pg-ha-62380f9c-no-violation"}
```
morgante, line 31: Are you expecting this fixture to raise a violation? If so, it should *not* be called "with_correctkey" and you *definitely* should not expect a resource called "no-violation" to raise a violation. Either your naming or logic is wrong.

morgante (changes requested):

validator/postgresql_log_checkpoints_test.rego
palani-ram-google-partner: @morgante This case(postgresql with correct key) don't raise a violation in our case. So I have removed the test case since it's not needed. Violated test cases alone is attached. Hope this is Ok. Please review it.

morgante (changes requested):
Please provide a test case which *doesn't violate*.

palani-ram-google-partner:
@morgante Added the test case with correct naming.

validator/postgresql_log_checkpoints_test.rego
```diff
@@ -0,0 +1,51 @@
...
+import data.test.fixtures.postgresql_log_checkpoints.assets.required as fixture_no_violation
+import data.test.fixtures.postgresql_log_checkpoints.assets.violations as fixture_violation
+
+import data.test.fixtures.postgresql_log_checkpoints.constraints as fixture_constraint
+
+template_name := "GCPPostgreSQLCheckpointsConstraintV1"
+
+#1. postgresql with correct key
+test_postgresql_log_checkpoints_required_databaseflag {
+	expected_resource_names := {"//cloudsql.googleapis.com/projects/my-test-project/instances/tf-pg-ha-62380f9c-required"}
```
morgante, line 31: Unfortunately, this is *still* wrong.
check_test_violations_count is asserting how many violations a particular fixture (data set) should raise. According to the naming of your fixture, it should *NOT* be violating.

Yet it *is* violating, since you:
(a) Are expecting a violation (count = 1).
(b) Have the name of the supposedly correct resource in the `expected_resource_names`. `expected_resource_names` is the list of resources you expect *to raise a violation*. A correct resource should never be found here.

It seems like your logic and tests are likely *both* broken.

Please read through the full test utils and look through other tests. It's important that you get this right.

morgante (changes requested):

PR #412, opened 2021-11-02
fix typos

e-koma:
fixed typos errors, there are many typo mistakes in the policy library.


morgante (approved):
Thanks for cleaning these up!

PR #415, opened 2021-11-17
Remove PodSecurityPolicy requirement for gcp_gke_restrict_pod_traffic_v1

jacks-reid:
This change removes the `pod_security_policy_config_enabled == true` requirement in the `gcp_gke_restrict_pod_traffic_v1.yaml` constraint template.

I believe the PodSecurityPolicy requirement should be removed because of its [recent deprecation in Kubernetes v1.21](https://kubernetes.io/blog/2021/04/06/podsecuritypolicy-deprecation-past-present-and-future/) and that users will more closely associate this policy with the enforcement of network policies.

To accomplish this:
- `pod_security_policy_config_enabled` was dropped from the `check_all_enabled` rule in the source Rego
- the `pod_security_policy_config_enabled` rule block was dropped altogether
- the constraint template has been updated with the new source Rego
- PSP-oriented test cases in `asset.json` were dropped and the Rego test file was updated to account for their removal


morgante (changes requested):
Since this is a substantial change to the template logic, we should probably make it a 'v2' policy. Can you archive the v1 policy and make a new v2 policy?

jacks-reid:
Makes sense.

- Created new `policies/templates/gcp_gke_restrict_pod_traffic_v2.yaml`
- Moved v1 template into `policies/templates/legacy`
- Adjusted `GCPGKERestrictPodTrafficConstraintV1` to V2 in `samples`
- Moved v1 sample into `samples/legacy/gke_restrict_pod_traffic_v1.yaml`
- Updated `validator/` Rego to use `GCPGKERestrictPodTrafficConstraintV2`

Are these all the changes you had in mind?

morgante:
@jacks-reid Yep, looks good. Thank you.

I need to get some tests fixed before I can give this a more thorough review though.

morgante (approved):

PR #417, opened 2021-11-20
fix: update doc generation to work with kpt v1

morgante:


melinath (approved):
I don't know enough to evaluate whether this will fix a particular issue but it seems reasonable.

PR #418, opened 2021-12-06
fix: Handle expiration dates in the far future 

morgante:
Adds a workaround for https://github.com/open-policy-agent/opa/issues/4098


g-awmalik (approved):
LGTM

PR #422, opened 2022-02-04
chore:updates the readme with the latest kpt commands

g-awmalik:
Updates the README.md:
- the kpt link to point to kpt.dev
- `kpt fn run` replaced with `kpt fn eval`

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,15 +6,15 @@ This repo contains a library of constraint templates and sample constraints.
 For information on setting up Config Validator to secure your environment, see the [User Guide](./docs/user_guide.md).
 
 ## Initializing a policy library
-You can easily set up a new (local) policy library by downloading a [bundle](./docs/index.md#policy-bundles) using [kpt](https://googlecontainertools.github.io/kpt/).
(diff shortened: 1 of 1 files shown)


g-awmalik:
@morgante - please review

policies/constraints/README.md
```diff
@@ -1 +0,0 @@
-This directory should contain the constraint .yaml files used in your environment. You can find sample constraints under https://github.com/GoogleCloudPlatform/policy-library/tree/master/samples.
```
morgante, line 1: Why is this being removed? It keeps the directory and is a helpful pointer IMO.

g-awmalik: If you run the bundler and sink it to the `policies\constraints` folder as mentioned in the README.md, it fails since the directory is already there. We could keep this and add a subfolder to `policies\constraints` as `policies\constraints\$BUNDLE`. That should work.

morgante: That would probably be better.

g-awmalik: Acked.

PR #423, opened 2022-02-08
feat:adds feature to generate a flat i.e. folder-less bundle

g-awmalik:
- Updates the kpt commands in the top-level readme to match kptv1
- Adds a flat file export flag that allows for constraints to be created without their source dir folder structure.


g-awmalik:
@morgante - can you please review this? This addresses the [earlier PR](https://github.com/GoogleCloudPlatform/policy-library/pull/422) as well which I just closed.

PR #427, opened 2022-04-22
docs: update user guide to use gcloud beta terraform vet

iyabchen:


docs/user_guide.md
```diff
@@ -9,7 +9,7 @@
   * [Get started with the Policy Library repository](#get-started-with-the-policy-library-repository)
   * [Instantiate constraints](#instantiate-constraints)
 * [How to use Terraform Validator](#how-to-use-terraform-validator)
```
melinath, line 11: As long as we're making these changes - it could be nice to update this as well. Possibly "Validate policies" or "Add policy validation" to match https://cloud.google.com/docs/terraform/policy-validation

iyabchen: Should we just say "refer to https://cloud.google.com/docs/terraform/policy-validation/validate-policies" to replace this whole section?

melinath: actually yeah that would probably be good. It doesn't seem like there's any other content in this section that wouldn't be covered.

iyabchen: Done.

melinath (approved):

g-awmalik (approved):
LGTM

PR #429, opened 2022-08-03
Update to ancestries & excluded ancestries

melinath:
b/241296218

This is now supported in gcloud terraform vet.

### DIFF
--- a/samples/gke_enable_logging.yaml
+++ b/samples/gke_enable_logging.yaml
@@ -21,7 +21,7 @@ metadata:
 spec:
   severity: high
   match:
-    target: # {"$ref":"#/definitions/io.k8s.cli.setters.target"}
+    ancestries:
     - "organizations/**"
   parameters:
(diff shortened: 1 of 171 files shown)


samples/gke_enable_logging.yaml
```diff
@@ -21,7 +21,7 @@ metadata:
 spec:
   severity: high
   match:
-    target: # {"$ref":"#/definitions/io.k8s.cli.setters.target"}
+    ancestries:
```
g-awmalik, line 24: nit: is there a reason some ancestries/targets are defined as indented values vs. others that are abbreviated flow collections like `["org/**"]`

g-awmalik (approved):

samples/gke_enable_logging.yaml
melinath: not that I'm aware of; it predates me. I believe it's the same from the yaml perspective?

PR #430, opened 2022-08-11
fix: update constraint for vpc flow logs to check on new field

### LINKED ISSUE #414: google_compute_subnetwork enable_flow_logs deprecated in google terraform 3.0.0 and GCPNetworkEnableFlowLogsConstraintV1 broken
I am applying the the https://github.com/terraform-google-modules/terraform-example-foundation for my company and used this policy library but encountered a problem when validating the 3-networks stage. We are using the Cloud Build instructions so we never customized terraform or google plugin versions at all. We are using most, if not all the policies from this repo because we need the added security for compliance reasons. The error we are getting is:

```
Constraint GCPNetworkEnableFlowLogsConstraintV1.enable_network_flow_logs on resource //compute.googleapis.com/projects/prj-c-base-net-hub-32b0/regions/us-central1/subnetworks/sb-c-shared-base-hub-us-central1: Flow logs are disabled in subnetwork //compute.googleapis.com/projects/prj-c-base-net-hub-32b0/regions/us-central1/subnetworks/sb-c-shared-base-hub-us-central1.
```

The issue is that subnet has logging enabled so it shouldn't be failing validation:

```
# module.base_shared_vpc[0].module.main.module.subnets.google_compute_subnetwork.subnetwork["us-central1/sb-c-shared-base-hub-us-central1"] will be created
+ resource "google_compute_subnetwork" "subnetwork" {
    + creation_timestamp         = (known after apply)
    + description                = "Base network hub subnet for us-central1"
    + external_ipv6_prefix       = (known after apply)
    + fingerprint                = (known after apply)
    + gateway_address            = (known after apply)
    + id                         = (known after apply)
    + ip_cidr_range              = "10.0.0.0/24"
    + ipv6_cidr_range            = (known after apply)
    + name                       = "sb-c-shared-base-hub-us-central1"
    + network                    = "vpc-c-shared-base-hub"
    + private_ip_google_access   = true
    + private_ipv6_google_access = (known after apply)
    + project                    = "prj-c-base-net-hub-32b0"
    + purpose                    = (known after apply)
    + region                     = "us-central1"
    + secondary_ip_range         = []
    + self_link                  = (known after apply)
    + stack_type                 = (known after apply)

    + log_config {
        + aggregation_interval = "INTERVAL_5_SEC"
        + filter_expr          = "true"
        + flow_sampling        = 0.5
        + metadata             = "INCLUDE_ALL_METADATA"
      }
  }
```

My best guess is the line https://github.com/GoogleCloudPlatform/policy-library/blob/master/policies/templates/gcp_network_enable_flow_logs_v1.yaml#L62 is reading `enableFlowLogs` from the resource data but the enable_flow_logs field is deprecated according to https://registry.terraform.io/providers/hashicorp/google/latest/docs/guides/version_3_upgrade#enable_flow_logs-is-now-removed

> enable_flow_logs has been removed and should be replaced by the log_config block with configurations for flow logging. Enablement of flow logs is now controlled by whether log_config is defined or not instead of by the enable_flow_logs variable. Users with enable_flow_logs = false only need to remove the field.

I assume there is a way to create my own GCPNetworkEnableFlowLogsConstraintV2 to fix this or more logic to check plugin versions, but I am new to terraform-validator and policies and these google golang libraries so it isn't obvious how to update the rego to fix it.

mauro-cit:
FIxes #414 

This PR main idea is to consider both versions (old and new) of Flow Logs parameters on the Subnetwork definition:
- Old version - before google provider 3.0: enable_flow_logs
- New version - after google provider 3.0: log_config

If both of them are false, the validation fails.
Some test data were added and tested on it's data test file validator/test/fixtures/network_enable_flow_logs/assets/data.json

**Manual tests:**
We tried to test it manually with a subnetwork resource definition using both versions of google provider (one where it uses the old field and one that uses the new field) but we faced a problem with the old version. 

When we used the validation with `gcloud beta terraform vet` and also `terraform validator` , apparently during the process to convert the Plan json file to CAI, it is always converting to the new model which have the logConfig instead. Even with the old versions with terraform-validator the it keeps converting to a CAI with logConfig. 

So we weren't able to test it manually with the old parameter after all.

### DIFF
--- a/policies/templates/gcp_network_enable_flow_logs_v1.yaml
+++ b/policies/templates/gcp_network_enable_flow_logs_v1.yaml
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
-# Ensure flow logs are enabled
+# Ensure flow logs are enabled 
 # https://cloud.google.com/vpc/docs/using-flow-logs
 
 apiVersion: templates.gatekeeper.sh/v1alpha1
@@ -45,11 +45,11 @@ spec:
             # See the License for the specific language governing permissions and
             # limitations under the License.
             #
-
+            
             package templates.gcp.GCPNetworkEnableFlowLogsConstraintV1
-
+            
             import data.validator.gcp.lib as lib
-
+            
             deny[{
             	"msg": message,
             	"details": metadata,
@@ -57,15 +57,15 @@ spec:
             	constraint := input.constraint
             	asset := input.asset
             	asset.asset_type == "compute.googleapis.com/Subnetwork"
-
+            
             	network := asset.resource.data
             	legacy_enable_flow_logs := lib.get_default(network, "enableFlowLogs", false)
             	log_config := lib.get_default(network, "logConfig", {})
             	log_config_enable_flow_logs := lib.get_default(log_config, "enable", false)
-
+            
             	log_config_enable_flow_logs != true
             	legacy_enable_flow_logs != true
-
+            
             	message := sprintf("Flow logs are disabled in subnetwork %v.", [asset.name])
             	metadata := {"resource": asset.name}
             }
--- a/policies/templates/gcp_network_enable_flow_logs_v1.yaml
+++ b/policies/templates/gcp_network_enable_flow_logs_v1.yaml
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
-# Ensure flow logs are enabled 
+# Ensure flow logs are enabled
 # https://cloud.google.com/vpc/docs/using-flow-logs
 
 apiVersion: templates.gatekeeper.sh/v1alpha1
@@ -45,11 +45,11 @@ spec:
             # See the License for the specific language governing permissions and
             # limitations under the License.
             #
-            
+
             package templates.gcp.GCPNetworkEnableFlowLogsConstraintV1
-            
+
             import data.validator.gcp.lib as lib
-            
+
             deny[{
             	"msg": message,
             	"details": metadata,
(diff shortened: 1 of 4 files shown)


bharathkkb:
@g-awmalik Not urgent, could you PTAL why the tests are failing? Also aren't these triggers in our CI/CD project, I don't have access to it?

g-awmalik:
@bharathkkb - these test run in the config validator project which i don't have access to either. @melinath - do you have access to the int-test project?

melinath:
I have access - the failure message is:

```
diff -r '--exclude=.git' /workspace/policies/templates/gcp_network_enable_flow_logs_v1.yaml /tmp/tmp.3GtBKekLGW/policies/templates/gcp_network_enable_flow_logs_v1.yaml
48c48
< 
---
>             
50c50
< 
---
>             
52c52
< 
---
>             
60c60
< 
---
>             
65c65
< 
---
>             
68c68
< 
---
>      
```

mauro-cit:
The build was fixed. @melinath @g-awmalik could you PTAL?

policies/templates/gcp_network_enable_flow_logs_v1.yaml
```diff
@@ -59,8 +59,12 @@ spec:
             	asset.asset_type == "compute.googleapis.com/Subnetwork"
             
             	network := asset.resource.data
-            	enable_flow_logs := lib.get_default(network, "enableFlowLogs", false)
-            	enable_flow_logs == false
+            	legacy_enable_flow_logs := lib.get_default(network, "enableFlowLogs", false)
+            	log_config := lib.get_default(network, "logConfig", {})
+            	log_config_enable_flow_logs := lib.get_default(log_config, "enable", false)
```
g-awmalik, line 64: Based on the provider [documentation](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/compute_subnetwork#log_config) I believe that the presence of this block denotes that logging is enabled. I don't think there is an `enable` argument in its schema to begin with.

I believe just the presence of this block can be used to assert that logs are enabled.

validator/test/fixtures/network_enable_flow_logs/assets/data.json
```diff
@@ -44,5 +44,82 @@
...
+        "fingerprint": "gHOs7GmZw+Q=",
+        "gatewayAddress": "10.140.0.1",
+        "id": "7589039100605557012",
+        "ipCidrRange": "10.140.0.0/20",
+        "name": "default",
+        "network": "https://www.googleapis.com/compute/v1/projects/pso-cicd8/global/networks/default",
+        "privateIpGoogleAccess": true,
+        "enableFlowLogs": false,
+        "logConfig": {
+          "enable": false
```
g-awmalik, line 67: Don't think we need this based on my comment on the template/rego

g-awmalik (commented):
Thanks for the PR @mauro-cit and apologies for the delay. Please take a look at my comments.

policies/templates/gcp_network_enable_flow_logs_v1.yaml
mauro-cit: Hello @g-awmalik , so, this "enable" field is actually a calculated field, it's generated through the [CAI conversion proccess](https://github.com/GoogleCloudPlatform/terraform-validator/blob/main/converters/google/resources/compute_subnetwork.go#L261).

So, when we verify whether it's true, we are checking if `lib.get_default(network, "logConfig", {})` didn't returned an empty map.

g-awmalik: Ok thanks for clarifying.

g-awmalik (approved):

PR #436, opened 2023-06-13
fix: add VPC Flow logs exceptions for REGIONAL_MANAGED_PROXY and INTERNAL_HTTPS_LOAD_BALANCER

### LINKED ISSUE #435: Validation for VPC flow logs in a subnetwork does not consider exceptions listed in the documentation
In the documentation related to the creation of subnetworks

- https://cloud.google.com/compute/docs/reference/rest/v1/subnetworks/insert
- https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/compute_subnetwork.html#purpose

 there are two exceptions in which it is not possible to enable vpc flow logs in a sub network. These are subnetworks with the following purposes:

- REGIONAL_MANAGED_PROXY
- INTERNAL_HTTPS_LOAD_BALANCER

the validation that checks if VPC flow logs is enable:

- https://github.com/GoogleCloudPlatform/policy-library/blob/c3382cbe86e5bae529dac0ac21e86679b606475c/validator/network_enable_flow_logs.rego
- https://github.com/GoogleCloudPlatform/policy-library/blob/c3382cbe86e5bae529dac0ac21e86679b606475c/policies/templates/gcp_network_enable_flow_logs_v1.yaml

is not considering these two case. 

This will mark as invalid the creation of a subnetwork for a [Secure Web Proxy](https://cloud.google.com/secure-web-proxy/docs/initial-setup-steps#create_a_proxy_subnet)  or an [Internal HTTP(S) load balancer](https://cloud.google.com/load-balancing/docs/l7-internal/int-https-lb-tf-examples#with_a_mig_backend) among other cases.

daniel-cit:
Fixes #435

### DIFF
--- a/validator/network_enable_flow_logs.rego
+++ b/validator/network_enable_flow_logs.rego
@@ -35,7 +35,7 @@ deny[{
 	log_config_enable_flow_logs != true
 	legacy_enable_flow_logs != true
 	purpose != "REGIONAL_MANAGED_PROXY"
-	purpose != "INTERNAL_HTTPS_LOAD_BALANCE"
+	purpose != "INTERNAL_HTTPS_LOAD_BALANCER"
 
 	message := sprintf("Flow logs are disabled in subnetwork %v.", [asset.name])
 	metadata := {"resource": asset.name}
--- a/validator/network_enable_flow_logs.rego
+++ b/validator/network_enable_flow_logs.rego
@@ -30,9 +30,12 @@ deny[{
 	legacy_enable_flow_logs := lib.get_default(network, "enableFlowLogs", false)
 	log_config := lib.get_default(network, "logConfig", {})
 	log_config_enable_flow_logs := lib.get_default(log_config, "enable", false)
+	purpose := lib.get_default(network, "purpose", "PRIVATE")
 
 	log_config_enable_flow_logs != true
 	legacy_enable_flow_logs != true
+	purpose != "REGIONAL_MANAGED_PROXY"
+	purpose != "INTERNAL_HTTPS_LOAD_BALANCE"
 
 	message := sprintf("Flow logs are disabled in subnetwork %v.", [asset.name])
 	metadata := {"resource": asset.name}
--- a/policies/templates/gcp_network_enable_flow_logs_v1.yaml
+++ b/policies/templates/gcp_network_enable_flow_logs_v1.yaml
@@ -62,9 +62,12 @@ spec:
             	legacy_enable_flow_logs := lib.get_default(network, "enableFlowLogs", false)
             	log_config := lib.get_default(network, "logConfig", {})
             	log_config_enable_flow_logs := lib.get_default(log_config, "enable", false)
+            	purpose := lib.get_default(network, "purpose", "PRIVATE")
             
             	log_config_enable_flow_logs != true
             	legacy_enable_flow_logs != true
+            	purpose != "REGIONAL_MANAGED_PROXY"
+            	purpose != "INTERNAL_HTTPS_LOAD_BALANCE"
             
             	message := sprintf("Flow logs are disabled in subnetwork %v.", [asset.name])
             	metadata := {"resource": asset.name}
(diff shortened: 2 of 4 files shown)


validator/network_enable_flow_logs.rego
```diff
@@ -30,9 +30,12 @@ deny[{
 	legacy_enable_flow_logs := lib.get_default(network, "enableFlowLogs", false)
 	log_config := lib.get_default(network, "logConfig", {})
 	log_config_enable_flow_logs := lib.get_default(log_config, "enable", false)
+	purpose := lib.get_default(network, "purpose", "PRIVATE")
 
 	log_config_enable_flow_logs != true
 	legacy_enable_flow_logs != true
+	purpose != "REGIONAL_MANAGED_PROXY"
+	purpose != "INTERNAL_HTTPS_LOAD_BALANCE"
```
g-awmalik, line 38: Shouldn't this be an "==" comparison since the flow logs will be disable IF the purpose IS set to `REGIONAL_MANAGED_PROXY` or `INTERNAL_HTTPS_LOAD_BALANCER` 

Also, I think this should be ORed as:
`any([purpose == "REGIONAL_MANAGED_PROXY", purpose == "INTERNAL_HTTPS_LOAD_BALANCER"])`

g-awmalik (changes requested):
Thanks for the PR. Please take a look at my comment.

validator/network_enable_flow_logs.rego
g-awmalik: Actually this may be bit more nuanced because if we determine that the flow logs are already set to false, we don't need to check for the purpose anymore. We would only check the purpose IF the flow logs were set to true.

daniel-cit: Hi @g-awmalik thanks for the review.
My thoughts on the points you highlighted

1) The semantics of the validation  check is to trigger a violation if **the following condition is `true`**: 

```
log_config_enable_flow_logs != true
legacy_enable_flow_logs != true
```

which could be rewritten as: 

- `log_config_enable_flow_logs` is not set to `true` **AND** `legacy_enable_flow_logs` is not set to `true` => rise **Violation**

with the change we will have:

```
log_config_enable_flow_logs != true
legacy_enable_flow_logs != true
purpose != "REGIONAL_MANAGED_PROXY"
purpose != "INTERNAL_HTTPS_LOAD_BALANCE"
```

which could be rewritten as: 

- `log_config_enable_flow_logs` is not set to `true` **AND** `legacy_enable_flow_logs` is not set to `true`
**AND** `purpose` is not `REGIONAL_MANAGED_PROXY` **AND** `purpose` is not `INTERNAL_HTTPS_LOAD_BALANCE` => rise **Violation**


2) on the other topic, of  a more nuanced validation, considering the case of subnetwork having `REGIONAL_MANAGED_PROXY` as the purpose we would have two cases:

- vpc flow log is NOT enabled and purpose is `REGIONAL_MANAGED_PROXY` => **No violation** is raised, and the resource configuration is OK
- vpc flow log is enabled and purpose is `REGIONAL_MANAGED_PROXY` => No violation is raised BUT the resource configuration is NOT OK (because if the purpose is `REGIONAL_MANAGED_PROXY` it should not have a vpc flow logs configuration) BUT this is one of the validations of the API and will rise an error in apply. this is not security recommendation issue.

g-awmalik: In that case we're not completely shifting left the validation. I.e. for a configuration that has `purpose` set as `REGIONAL_MANAGED_PROXY` and has flow logs enabled, there will be no violation which raises question if adding a purpose check is adding any value?

daniel-cit: do you mean that we should use

```
purpose != "REGIONAL_MANAGED_PROXY"
purpose != "INTERNAL_HTTPS_LOAD_BALANCE"
log_config_enable_flow_logs != true
legacy_enable_flow_logs != true
```

g-awmalik: But this doesn't catch the scenario where purpose is `REGIONAL_MANAGED_PROXY` or `INTERNAL_HTTPS_LOAD_BALANCER` and has flow logs enabled. That won't raise a violation but will return and error at the API level.

I don't know if that's even possible with rego and that's why I'm suggesting we don't make this change since it doesn't solve the problem completely. I'll have to consume a few brain cells to see if this is even possible in rego.

g-awmalik (approved):

